PR #14, opened 2017-07-21
corrected minor typo

crowchirp:


kiss96803:
Thanks for your  PR, We checked and found the word is indeed wrong

kiss96803:
Thanks for your PR, We've checked and the word is indeed wrong!   : )

PR #159, opened 2018-07-16
新增GBase8t的CAP

wwwyyt000:


yuleyule66:
Thanks for your PR, Gbase database may not be used by most people, we need to consider the needs and compatibility of most people, so I do not recommend adding this feature to the list of official support. If possible, are you willing to exist as an extended project and continue to maintain it?

wwwyyt000:
yes




At 2018-07-17 09:16:38, "Savorboard" <notifications@github.com> wrote:


Thanks for your PR, Gbase database may not be used by most people, we need to consider the needs and compatibility of most people, so I do not recommend adding this feature to the list of official support. If possible, are you willing to exist as an extended project and continue to maintain it?

PR #167, opened 2018-07-23
Abstract MongoTransaction to make the usage more comfortable

kexxxfeng:
Improve the usage like this:
```c#
using (var trans = await _mongoTransaction.BegeinAsync())
{
    repository.InsertOne(trans.GetSession(), someValue);
    await _capPublisher.PublishWithMongoAsync("sample.rabbitmq.mongodb", DateTime.Now, trans);
}
```


kexxxfeng:
There was a bug. I will fix it soon.

kexxxfeng:
Because the default collection name was updated so that `counter` doesn't work after `git pull`. Dropped the database fixed the bug. So nothing need to be updated.

PR #177, opened 2018-08-07
Mock Cap

kexxxfeng:
`AddMockCap` is for tests which need mock cap.


kexxxfeng:
Usage
```c#
services.AddMockCap(x =>
{
    x.UseMockEntityFramework();
});
```
It's convenient for integration tests which doesn't take care about 'event bus'.

yuleyule66:
@kexxxfeng  thanks your PR, I think we can't add mock code to the main code, If we really need it, maybe we need use conditional compilation Isolate the mock code.

kexxxfeng:
Ok

PR #212, opened 2018-09-25
Ensure temporary RabbitMQ disconnection can be recovered

CleanCoder:
In Production, we suffered the following RabbitMQ error, and resolved by this change.

The AMQP operation was interrupted: AMQP close-reason, initiated by Library, code=0, text="End of stream", classId=0, methodId=0, cause=System.IO.IOException: Unable to write data to the transport connection: An existing connection was forcibly closed by the remote host. ---> System.Net.Sockets.SocketException: An existing connection was forcibly closed by the remote host
   at System.Net.Sockets.NetworkStream.Write(Byte[] buffer, Int32 offset, Int32 size)
   --- End of inner exception stack trace ---
   at System.Net.Sockets.NetworkStream.Write(Byte[] buffer, Int32 offset, Int32 size)
   at System.IO.BufferedStream.Flush()
   at RabbitMQ.Client.Impl.SocketFrameHandler_0_9.WriteFrame(Frame frame)
   at RabbitMQ.Client.Impl.ConnectionBase.HeartbeatWriteLoop()


src/DotNetCore.CAP.RabbitMQ/IConnectionChannelPool.Default.cs
```diff
@@ -98,7 +98,10 @@ public virtual IModel Rent()
 
                 Debug.Assert(_count >= 0);
 
-                return model;
+                if (model.IsOpen && model.CloseReason == null)
```
yang-xiaodong: If we have already detected the state of the connection at the time of the return, I think it is unnecessary to check again

src/DotNetCore.CAP.RabbitMQ/IConnectionChannelPool.Default.cs
```diff
@@ -108,7 +111,7 @@ public virtual IModel Rent()
 
         public virtual bool Return(IModel connection)
         {
-            if (Interlocked.Increment(ref _count) <= _maxSize)
+             if (connection.IsOpen && Interlocked.Increment(ref _count) <= _maxSize)
```
yang-xiaodong: I think we need add the check of the model connection to the `IConnectionChannelPool.Return(IModel connection)` method , because  the `public virtual bool Return(IModel connection)` can be override by user, It might be missed.

src/DotNetCore.CAP.RabbitMQ/IConnectionChannelPool.Default.cs
CleanCoder: The connection can be disconnected at any time even it has been returned. It should better add a safe check here.

src/DotNetCore.CAP.RabbitMQ/IConnectionChannelPool.Default.cs
CleanCoder: Not sure what's your mean of adding dd the check of the model connection to the IConnectionChannelPool.Return(IModel connection), it's an interface definition. 
For your concern, I think the developer who override the method should take this responsibility to ensure not return the unhealthy connection like what we do here :).

yang-xiaodong: we need keep the program robustness,  If user return a bad connection ,we should not add it to the connection pool,  so I think this check is necessary

CleanCoder: Please ignore my PR. Just noticed the latest RabbitMQ.Client has already support auto recovery  feature and by default it turns on. Please refer to the ConnectionFactyory class
public bool AutomaticRecoveryEnabled { get; set;} = true;

We used the old version (V3.x) due to compatibility. 
Sorry for the trouble and thanks for your time for reviewing my PR.

yang-xiaodong:
Base on the discuss, the client driver already has automatic recovery,  I will close this PR. Thanks again.  :)

PR #300, opened 2019-03-16
Prefixed arguments in RabbitMQConsumerClient.cs

xXjoakinXx:


yang-xiaodong:
CI not passed, can you fix it？

xXjoakinXx:
Error message:
"Named argument specifications must appear after all fixed arguments have been specified. Please use language version 7.2 or greater to allow non-trailing named arguments. "

`_channel.QueueDeclare(_queueName, passive: true, durable: false, autoDelete: false, arguments);`

The problem comes with the version of the C# language that is being used, it is lower than 7.2.

Apparently in old versions it is not allowed to establish names in the arguments as long as these are established last in the declaration of the method. I have updated to version 7.2 and it has worked for me.

The decision to update the language version is in your hands. 😃

yang-xiaodong:
Add `arguments:` to lasted argument to solved ,I have fixed it . thanks

PR #324, opened 2019-04-19
Handle messages retrieval failure

stubbst:
When the `IStorageConnection` service is down, `connection.GetPublishedMessagesOfNeedRetry()` and `connection.GetReceivedMessagesOfNeedRetry()` both throw an exception.

The wait operation in `ProcessAsync` is therefore never executed.

We therefore end up in an infinite loop of immediate retries until the `IStorageConnection` service is back up. This can be seen in the logs where we have hundreds of "Processor 'DotNetCore.CAP.Processor.NeedRetryMessageProcessor' failed. Retrying..." entries per second.

We add a try/catch around the message retrieval operation and log the error. The next messages retrieval will therefore be executed `options.FailedRetryInterval` seconds later.

### DIFF
--- a/src/DotNetCore.CAP/Processor/IProcessor.NeedRetry.cs
+++ b/src/DotNetCore.CAP/Processor/IProcessor.NeedRetry.cs
@@ -48,7 +48,7 @@ namespace DotNetCore.CAP.Processor
 
         private async Task ProcessPublishedAsync(IStorageConnection connection, ProcessingContext context)
         {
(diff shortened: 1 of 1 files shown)


yang-xiaodong:
Thanks for your PR

It seems that you have not finished yet, I can't find where to call `GetSafelyAsync`

stubbst:
Hi. Sorry about that. I forgot to retake two crucial lines from the modifications I made in my project where I overrode some of the classes 😕.

PR #335, opened 2019-05-01
fix NullReferenceException

leisurelyclouds:
在启用 CapBeforePublish 后 发送消息时会导致空引用异常

when DiagnosticListener enabled, publish message will cause null reference exception

Call Stack:
TracingHeaders.GetEnumerator
TracingHeaders.IEnumerable.GetEnumerator
ToJObject
Helper.AddTracingHeaderProperty
BasePublishMessageSender.SendWithoutRetryAsync
BasePublishMessageSender.SendAsync
NeedRetryMessageProcessor.ProcessPublishedAsync
NeedRetryMessageProcessor.ProcessAsync
BasePublishMessageSender.SendWithoutRetryAsync
BasePublishMessageSender.SendAsync

### DIFF
--- a/src/DotNetCore.CAP/Diagnostics/TracingHeaders.cs
+++ b/src/DotNetCore.CAP/Diagnostics/TracingHeaders.cs
@@ -9,7 +9,7 @@ namespace DotNetCore.CAP.Diagnostics
 {
     public class TracingHeaders : IEnumerable<KeyValuePair<string, string>>
     {
-        private List<KeyValuePair<string, string>> _dataStore;
+        private List<KeyValuePair<string, string>> _dataStore = new List<KeyValuePair<string, string>>();
 
         public IEnumerator<KeyValuePair<string, string>> GetEnumerator()
         {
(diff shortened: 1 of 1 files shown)


src/DotNetCore.CAP/Diagnostics/TracingHeaders.cs
```diff
@@ -23,11 +23,6 @@ IEnumerator IEnumerable.GetEnumerator()
 
         public void Add(string name, string value)
         {
-            if (_dataStore == null)
```
yang-xiaodong: Can you tell me why it didn't work?

leisurelyclouds:
设置docker-compose 为启动项
访问http://localhost:5678/ef/transaction 能重现bug

https://github.com/leisurelyclouds/CAP/tree/PR335Bug
set  docker-compose as startup project
visit http://localhost:5678/ef/transaction will reproduce null reference exception

https://github.com/leisurelyclouds/CAP/blob/37b2313bdb45933f3e5dfc35cbd2ee7a707ea720/samples/Sample.RabbitMQ.MySql/Startup.cs#L74
https://github.com/opentracing-contrib/csharp-netcore/blob/f47503f6277b35f0be7ec6db871de0b2285dd2b6/src/OpenTracing.Contrib.NetCore/Internal/DiagnosticManager.cs#L53

yang-xiaodong:
I got the exception, thanks for your PR

PR #357, opened 2019-07-01
FindConsumersFromControllerTypes:  Search for controllers from AppDomaim.

hetaoos:
Some controllers are outside of `Entry Assembly`. 

For example, some controllers are dynamically loaded using `_mvcBuilder.AddApplicationPart(assembly)`. 

Of course, you can also use [IActionDescriptorCollectionProvider](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.infrastructure.iactiondescriptorcollectionprovider?view=aspnetcore-2.2) to get all the controller information, but this needs to reference [Microsoft.AspNetCore.Mvc.Core](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Core/).

### DIFF
--- a/src/DotNetCore.CAP/IConsumerServiceSelector.Default.cs
+++ b/src/DotNetCore.CAP/IConsumerServiceSelector.Default.cs
@@ -97,16 +97,17 @@ namespace DotNetCore.CAP
         {
             var executorDescriptorList = new List<ConsumerExecutorDescriptor>();
 
-            var types = Assembly.GetEntryAssembly().ExportedTypes;
-            foreach (var type in types)
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies().Where(o => !o.IsDynamic))
             {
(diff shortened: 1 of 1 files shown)


yang-xiaodong:
Thank you for your PR.

We thought about this at the beginning of the design. Since most users don't put the Controller in an external assembly, we didn't look for all the assemblies to find the Controller, which would significantly reduce the efficiency of the search.

Instead, we provide a way for the user to customize the Controller's lookup algorithm, so we'll mark `FindConsumersFromInterfaceTypes` as `virtual ` so that the user can rewrite on demand.

hetaoos:
Thank you for your reply.

Some developers may not know to rewrite the method if they don't read the source code. 
Maybe you can remind the developer in the documentation.

yang-xiaodong:
Sure, I will mark this point in the document

PR #367, opened 2019-07-17
modify single message sending process

lukazh:
ensure asynchronized processing and exception get caught

### DIFF
--- a/src/DotNetCore.CAP/Processor/IDispatcher.Default.cs
+++ b/src/DotNetCore.CAP/Processor/IDispatcher.Default.cs
@@ -59,14 +59,17 @@ namespace DotNetCore.CAP.Processor
                 {
                     if (_publishedMessageQueue.TryTake(out var message, 100, _cts.Token))
                     {
-                        try
+                        Task.Run(async () =>
                         {
-                            _sender.SendAsync(message);
-                        }
-                        catch (Exception ex)
-                        {
-                            _logger.LogError(ex, $"An exception occurred when sending a message to the MQ. Topic:{message.Name}, Id:{message.Id}");
-                        }
+                            try
+                            {
+                                await _sender.SendAsync(message);
+                            }
+                            catch (Exception ex)
+                            {
+                                _logger.LogError(ex, $"An exception occurred when sending a message to the MQ. Topic:{message.Name}, Id:{message.Id}");
+                            }
+                        });
                     }
                 }
             }


yang-xiaodong:
Asynchronous merge with #261 ,  Can't you receive the exception log?

lukazh:
#261 was reverted by 4a79956. Is there any problem with it?

yang-xiaodong:
That is we moved the  `Task.Run`  forward of [4a79956](https://github.com/dotnetcore/CAP/commit/4a79956912382d4fd23977c2d6b2479cfe79cca9). 

What is the reason of remove `Task.Run`?

lukazh:
Sorry I don't get your point. This PR is to **add** `Task.Run`, not **remove**. It was introduced in #261 and then removed in 4a79956.

yang-xiaodong:
Sorry, I miss it. 

This commit([4a79956](https://github.com/dotnetcore/CAP/commit/4a79956912382d4fd23977c2d6b2479cfe79cca9) ) is to fix #331, which contains thread safety issues. 

In v2.6.0 we adjusted `ICapPublisher` to singleton. Maybe we can use `Task.Run` to improve the sending speed, but he needs to do more testing. Would you like to attach a test report about this PR?

lukazh:
I did a simple test to find that most of the time in a whole publishing process is used to interact with database, so in normal usage of CAP the bottleneck should never be the "publish to MQ" part.

On the other hand, regarding to the "publish to MQ" itself, taking `Task.Run` for parallel processing will significantly decrease the time comsumption. `Task.Run` a `Action` is much faster than calling an async method without `await` because it will take some time to do some sync work before yielding control to the caller.

Last but not least, this PR is also trying to fix the problem that exceptions thrown in `IPublishMessageSender.SendAsync` may not get caught because the method is `async`.

PS: this PR may have something to do with #303 on the publish/consume order problem.

yang-xiaodong:
Hello @lukazh ,

Before you replied, I did some refactoring of the code, and do some tests about the performance.
I found the `Task.Run` can improve performance without side effects, so I fixed it in commit https://github.com/dotnetcore/CAP/commit/f664b62831a2f2db5c30f1dfa63c8f3a92a6155b#diff-5927512797ef2b8552052b0ea63e805eR64 

Anyway, thanks the PR!

lukazh:
Hi @yang-xiaodong , sorry for the delay.

Good to see the usage of `Task.Run`, but the **Exception won't get caught** problem still exists. `try...catch` should be inside the `Task.Run`. Throw an exception in `BasePublishMessageSender.SendAsync` to prove this.

yang-xiaodong:
Oh...  yes, I missed, I'll fix it for a moment.
Thanks.

yang-xiaodong:
Fixed in commit d1c26b7

PR #397, opened 2019-09-13
Add kafka key to publishing

ajaffie:
This PR adds the ability to send a string as a key in Kafka. The interface was modified to include the key as an optional parameter, which all message queue implementations ignore except Kafka. DB schemas were also updated to add the column if it doesn't exist.


yang-xiaodong:
This request has been implemented in the upcoming v3.0 release, we using topic as the key

dameeks:
@yang-xiaodong I do not see where to provide the message key in 3.0. We need to use the message keys to guarantee the order of certain messages within a topic. Where does CAP 3.0 make use of the message keys?

Issue #404, opened 2019-09-30
.net core 3.0 dashboard can not load js and css files


SlientHome:
For NetCore 3.0，Dashboard Can't Load Js And Css！

MahmoudH96:
Me too, I am getting 404 for the following URLs:
https://localhost:44311/cap/js260
https://localhost:44311/cap/css260

HuUncle:
I have throw this exception too

HuUncle:
add this code 

![TIM图片20191118153320](https://user-images.githubusercontent.com/19486270/69032984-d6f38d80-0a18-11ea-889d-9800f036aa96.png)

yang-xiaodong:
v3.0 nuget package is released!

PR #407, opened 2019-10-09
Fixed Net Core 3.0，Dashboard Can't Load Js And Css.

yrinleung:
Fixed Net Core 3.0，Dashboard Can't Load Js And Css.(#404)

### DIFF
--- a/src/DotNetCore.CAP/Dashboard/CombinedResourceDispatcher.cs
+++ b/src/DotNetCore.CAP/Dashboard/CombinedResourceDispatcher.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Reflection;
+using System.Threading.Tasks;
 
 namespace DotNetCore.CAP.Dashboard
 {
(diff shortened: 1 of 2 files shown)


yang-xiaodong:
Thanks for your PR, Do you konw why it works when update to async ?

yrinleung:
我查看到报错在[HttpResponseStream](https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/Core/src/Internal/Http/HttpResponseStream.cs)类中：
```
public override void Write(byte[] buffer, int offset, int count)
        {
            if (!_bodyControl.AllowSynchronousIO)
            {
                throw new InvalidOperationException(CoreStrings.SynchronousWritesDisallowed);
            }

            WriteAsync(buffer, offset, count, default).GetAwaiter().GetResult();
        }
```
而[AllowSynchronousIO]([https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/Core/src/KestrelServerOptions.cs) 在Net Core 2.2或以下版本默认是true，在Net Core 3.0调整为false。

可以查看官方调整的[PR](https://github.com/aspnet/AspNetCore/commit/93a24b03bbda1aa0ab9b553a50b70dd36d554934)

bao2314483:
我也遇到了,问题解决了吗

Issue #460, opened 2019-12-20
Rabbitmq message not sent


wangronghua:
This method successfully saved the record in the published database, but rabbitmq did not receive the message until 4 minutes later
``` c#
        using (var connection = new MySqlConnection(ConnectionString))
        {
            using (var transaction = connection.BeginTransaction(_capBus, autoCommit: true))
            {
                //Business code
                _capBus.Publish("xxx.services.show.time", DateTime.Now);
            }
        }
```
In addition, in the case of global transaction, the global transaction has not been committed, but rabbitmq has been sent.
``` c#
using _transactionScope = new TransactionScope(
                    TransactionScopeOption.Required,
                    new TransactionOptions { IsolationLevel = IsolationLevel.ReadCommitted })
{
                //Business code
                _capBus.Publish("xxx.services.show.time", DateTime.Now);
}
```

yang-xiaodong:
https://zhuanlan.zhihu.com/p/25795393

PR #461, opened 2019-12-20
 #460 add the function of sending message by manual control

wangronghua:
Add the function of manually controlling the sending of messages. By default, the existing automatic sending of messages does not need to change the previous code, so as to solve the problem in  #460

### DIFF
--- a/src/DotNetCore.CAP.RabbitMQ/DotNetCore.CAP.RabbitMQ.csproj
+++ b/src/DotNetCore.CAP.RabbitMQ/DotNetCore.CAP.RabbitMQ.csproj
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="RabbitMQ.Client" Version="5.1.2" />
+    <PackageReference Include="RabbitMQ.Client" Version="5.1.0" />
   </ItemGroup>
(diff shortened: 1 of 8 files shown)


wangronghua:
``` c#
try
{
            using (var connection = new SqlConnection(SingletonInstance<IMicroServiceEngine>.Instance.ConnectionString))
            {
                using (var transaction = connection.BeginTransaction(NotSingletonInstance<ICapPublisher>.Instance, autoCommit: true))
                {
                    //Business code
                    _capBus.Publish("xxx.services.show.time1", DateTime.Now, manuallySendMq:true);
                }
            }
            _capBus.ManuallySendMq();//If the business throws an error and the transaction is not committed, the message will not be sent
}
catch
{

}
```

wangronghua:
``` c#
try
{
            using (var connection = new SqlConnection(connectionString))
            {
                using (var transaction = connection.BeginTransaction(_capBus, autoCommit: true))
                {
                    //Business code
                    _capBus.Publish("xxx.services.show.time1", DateTime.Now, manuallySendMq:true);
                }
            }
            _capBus.ManuallySendMq();//If the business throws an error and the transaction is not committed, the message will not be sent
}
catch
{

}
```

PR #467, opened 2019-12-24
Improve build script

mzorec:
I've done some improvements in build script. 

I'am not sure how do you publish nuget packages. I am guessing you are publishing them manually so if you would like we could automate this to. There are alot of possibilites when build process publishes nuget packages. For example nuget package can be published only when you merge to master(or any other branch like "stable") and version is increased in version.props or you could use git version https://gitversion.net/docs/more-info/version-increments


mzorec:
Ah now I've noticed that you are using appveyor for publishing nuget packages. If you would prefere it in build script we can still change it.

yang-xiaodong:
Hello @mzorec 
Yes, the preview package is publishing manually on appveyor. The official package is automatically triggered at the github when publish release

mzorec:
Would you prefere that it stays this way or would you like that it is published through flubu build script on appveyor or any other CI?

yang-xiaodong:
Yes, I want keey the current way. 

Merry Christmas！

mzorec:
Thank you :)

Yes nuget publishing is as you had it before i haven't changed it abit.

PR #517, opened 2020-02-27
Add a group prefix

zhenlei520:
With the expansion of the project, after the basic function is extracted, the same service may need to subscribe to a message queue and specify to complete a certain function. Because the basic service is extracted into the basic package storage, the group name in the subscribed message queue Must be constants. It is convenient to complete the respective management by setting the prefix of the message queue of the service.
        Of course, the above are just examples for the problems I have encountered. Of course, you can add special expressions to the message queue routing by setting a prefix to achieve the same effect as Redis Prefix. If this function does not work for you, you can not set GroupPrefix, the last Group is the same as before,。
         I hope to pass the review, thank you very much

### DIFF
--- a/src/DotNetCore.CAP/CAP.Options.cs
+++ b/src/DotNetCore.CAP/CAP.Options.cs
@@ -24,6 +24,7 @@ namespace DotNetCore.CAP
             Extensions = new List<ICapOptionsExtension>();
             Version = "v1";
             DefaultGroup = "cap.queue." + Assembly.GetEntryAssembly()?.GetName().Name.ToLower();
+            GroupPrefix = "";
         }
 
         internal IList<ICapOptionsExtension> Extensions { get; }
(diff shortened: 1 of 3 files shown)


yang-xiaodong:
We do not want to achieve certain functions by always adding configuration items. In fact, we have been reducing the number of configuration items as much as possible. He will reduce the user's learning costs. We are cautious about adding new configuration items when it is not necessary. 

This you mentioned can be achieved in the project with some specifications and constraints. For example, in the Group constant to specify the prefix of a specific Team.

Anyway , Thanks for your PR!

zhenlei520:
I can understand your concerns. I now want to strip the underlying method of the project, where each service in the cache needs to be maintained independently, not using the same cache. I hope to make the basics through the message queue. The library performs different deletion caches by subscribing to the same route, and the Group requirement in CapSubscribe must be constant. If it is hard-coded in the project, each service cannot execute the message queue information once. How should this be handled?

zhenlei520:
What I can think of at the moment is to do something with Version, but it feels better without the prefix.

yang-xiaodong:
Depending on your business needs, you can do this by override the `SetSubscribeAttribute` method

PR #524, opened 2020-03-11
Removes line that added the kafka key to headers an extra time.

ajaffie:
This caused any messages published by CAP with a kafka key set to not be consumable by CAP.


ajaffie:
Hey @yang-xiaodong, would you be able to prioritize this PR?

PR #558, opened 2020-04-29
Add sqlserver test code

yimogit:


yang-xiaodong:
Thank you for your PR. In order to simplify the code, we removed the integration test of other databases except mysql in version 3.0.

PR #565, opened 2020-05-11
Fix : different methodInfo but get same executor

yrinleung:
Fix : different methodInfo but get same executor

### DIFF
--- a/src/DotNetCore.CAP/Internal/ISubscribeInvoker.Default.cs
+++ b/src/DotNetCore.CAP/Internal/ISubscribeInvoker.Default.cs
@@ -36,7 +36,7 @@ namespace DotNetCore.CAP.Internal
 
             _logger.LogDebug("Executing subscriber method : {0}", methodInfo.Name);
 
-            var executor = _executors.GetOrAdd(methodInfo.MetadataToken, x => ObjectMethodExecutor.Create(methodInfo, context.ConsumerDescriptor.ImplTypeInfo));
+            var executor = _executors.GetOrAdd(methodInfo.GetHashCode(), x => ObjectMethodExecutor.Create(methodInfo, context.ConsumerDescriptor.ImplTypeInfo));
 
             using (var scope = _serviceProvider.CreateScope())
             {


yang-xiaodong:
Hello, Thanks for your PR.

Please describe the scene and example of MetadataToken for methodInfo with the same values?

According to the explanation of MSDN:

https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.metadatatoken?view=netcore-3.1#System_Reflection_MemberInfo_MetadataToken

> MetadataToken
> A value which, in combination with Module, uniquely identifies a metadata element.

yrinleung:
```
    public interface IIntegrationEventHandler<TIntegrationEvent> 
    {
        Task Handle(TIntegrationEvent @event);
    }
    public class TestIntegrationEvent
    {
    }
    public class Test1IntegrationEvent
    {
    }


    var metadataToken = typeof(IIntegrationEventHandler<TestIntegrationEvent>).MetadataToken;
    var metadataToken1 = typeof(IIntegrationEventHandler<Test1IntegrationEvent>).MetadataToken;

    var bl = metadataToken == metadataToken1;
    Console.WriteLine($"metadataToken:{metadataToken}");
    Console.WriteLine($"metadataToken1:{metadataToken1}");
    Console.WriteLine($"bl:{bl}");
```
Result:
image

yrinleung:
scene:
> In the project, I rewrite “ConsumerServiceSelector"  to read “IIntegrationEventHandler&#60;TIntegrationEvent&#62;" convert "ConsumerExecutorDescriptor” ,so "metadatatoken" is the same

yang-xiaodong:
You're using the same object to printing the value

yrinleung:
Yes, but the subscription in the old project is based on generic subscription, so only be compatible
```
public interface IEventBus
{
        void Subscribe<T, TH>()
            where T : IIntegrationEvent
            where TH : IIntegrationEventHandler<T>;
}
```
In the current project, I rewrite "ISubscribeInvoker" to compatible

yang-xiaodong:
Hello, we don't think this is a bug that needs to be fixed or improved.

GetHashCode does not guarantee that different objects will return unique values.

You can implement ISubscribeInvoker to customize yours behavior

yang-xiaodong:
Hello, we don't think this is a bug that needs to be fixed or improved.

GetHashCode does not guarantee that different objects will return unique values.

You can implement ISubscribeInvoker to customize yours behavior

yrinleung:
OK , thank you

PR #575, opened 2020-05-23
update RabbitMQ.Client to 6.0.0

jerrytang67:


yang-xiaodong:
https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/CHANGELOG.md#changes-between-520-and-600

I still see a lot of optimization work in the 6.1 version, so I think that we are currently using 6.0 is not a good time.

Does CAP use Memory to improve performance? thoughts?

Maybe we need to wait until 6.1 is released and then consider switching to

jerrytang67:
ok, I publish this feature to my owner nuget, because other project refence RabblitMq.Client 6. If Cap.RabbitMq not use 6, will throw a Warning: cant't find some methord.

yang-xiaodong:
In the latest preview release, we have switched RabbitMQ.Client to 6.1

Issue #401, opened 2019-09-27
Is message delay supported?


NaturalDragon:

hu372197979:
No thanks,you need Hangfire!

yang-xiaodong:
The current version is not supported, and the custom header feature can be used in the next version to enable RabbitMQ to support it.

PR #578, opened 2020-06-01
add exchange type in consumer to support the delayed message

murdering:
### Feature: 
To support delayed rabbitmq messages, can extend Exchange Type to be configurable #401 

Pls see the commit logs,  I amend `RabbitMQOptions` to make `ExchangeType` as a settable property.
Then I tested in samples by sending out some delayed messages and received by consumer client. Also I start up one similar [java consumer client](https://www.jianshu.com/p/3bd861c4b479)。They have same delayed behavior as below.

### Test Result:
image

### Mysql records:
Publish:
{"Headers":{"x-delay":"6000","cap-msg-id":"1267395401949954048","cap-msg-name":"rk.delayed","cap-msg-type":"String","cap-senttime":"2020/6/1 17:59:45 +08:00","cap-corr-id":"1267395401949954048","cap-corr-seq":"0"},"Value":"this is a delayed message,current date: 2020/6/1 17:59:45"}

Consumer:
{"Headers":{"x-delay":"6000","cap-msg-id":"1267395401949954048","cap-msg-name":"rk.delayed","cap-msg-type":"String","cap-senttime":"2020/6/1 17:59:45 +08:00","cap-corr-id":"1267395401949954048","cap-corr-seq":"0","cap-msg-group":"queue.delayed.net.v1"},"Value":"this is a delayed message,current date: 2020/6/1 17:59:45"}

### DIFF
--- a/samples/Sample.ConsoleApp/EventSubscriber.cs
+++ b/samples/Sample.ConsoleApp/EventSubscriber.cs
@@ -1,5 +1,5 @@
-﻿using System;
-using DotNetCore.CAP;
+﻿using DotNetCore.CAP;
+using System;
 
 namespace Sample.ConsoleApp
 {
@@ -9,6 +9,17 @@ namespace Sample.ConsoleApp
         public void ShowTime(DateTime date)
         {
             Console.WriteLine(date);
+        }
+
+        /// <summary>
+        /// 延迟队列消费端
+        /// </summary>
+        /// <param name="senttime">发送时间</param>
+        [CapSubscribe("rk.delayed", Group = "queue.delayed.net")]
+        public void ShowTime(string senttime)
+        {
+            Console.WriteLine($"process time: {senttime}");
+            Console.WriteLine($"current time: {DateTime.Now}");
         }
     }
 }
--- a/samples/Sample.ConsoleApp/Program.cs
+++ b/samples/Sample.ConsoleApp/Program.cs
@@ -1,7 +1,7 @@
-﻿using System;
-using DotNetCore.CAP.Internal;
+﻿using DotNetCore.CAP.Internal;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using System;
 
 namespace Sample.ConsoleApp
 {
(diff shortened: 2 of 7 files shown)


yang-xiaodong:
Hello, thank you for your PR.

I think this feature  incompatible the roadmap of CAP. We hope that CAP can provide a consistent  external interface. RabbitMQ is only part of our transport supported, I don’t want to support non-universal features. On the contrary, I think that the delay feature should be to using the middleware like Quartz.

yang-xiaodong:
Hello, thank you for your PR.

I think this feature  incompatible the roadmap of CAP. We hope that CAP can provide a consistent  external interface. RabbitMQ is only part of our transport supported, I don’t want to support non-universal features. On the contrary, I think that the delay feature should be to using the middleware like Quartz.

murdering:
Hi, thanks for comment.

I see the rabbitmq exchange type always declaring as `topic `in CAP by a const variable. 
Actually, rabbitmq supports 4 basic types such as `Direct`, `Fanout`, `Topic` and `Headers`. 
Not sure why don't support the others by making it configurable?

Yes, I know the delayed message is not the common feature in many projects. 
But if the CAP project makes Exchange Type configurable, we can say it supports the various rabbitmq Exchange Type even the delayed type "x-delayed-message".

Appreciate to correct my misunderstanding. Thanks.

The core change is on "CAP.RabbiMQOptions.ExChangeType" settable.
image

yang-xiaodong:
CAP is not an encapsulated component of RabbitMQ. If we provide this kind of function, we must ensure that other Transport can also provide the same behavior, which can ensure that even if the producer and consumer use different Trasnports, there will be no inconsistent behavior.

murdering:
Hi, you may have a code glance of 3 files changes below in commit log。Ignore those in samples which is to set exchange type and delcare corresponding Exchange of RabbitMQ correctly.

1. `CAP.RabbiMQOptions.cs`
2. `ITransport.RabbitMQ.cs`
3. `RabbitMQConsumerClient.cs`

**First**, `RabbitMQOptions` is standalone class to set the RabbitMQ config, then I just open `ExChangeType` to be configurable and default value is still `topic`.

**Then**, change Exchange Declaration function invocation in RabbitMQ consumer client(file 2 and 3) to use the `ExChangeType` accordingly instead the const value before.
image

That's all the changes. I think this will not affect the other transport, and just edit in RabbitMQ specific.

Of course, CAP does not support the RabbitMQ only, but if it could support more the features, I think we will use it in more situations. Thx

Issue #577, opened 2020-05-28
Recommend to create a new independent extension project for efcore dependency


xiangxiren:
Currently, dapper and efcore are both referenced. If you use one of the other ORM frameworks in your project, you rely on three at the same time. Therefore, I recommend migrating the dependency on the efcore to a new extension project that can be referenced by the consumer separately. Even, I recommend removing the dependency on dapper and using native ado.net to store and read messages. Of course, it's just a suggestion.

yang-xiaodong:
Would you like to submit a PR that replaces dapper with ado.net?

yang-xiaodong:
New feature has been release with nuget package `3.1.0-preview-108719052`

PR #583, opened 2020-06-11
replace dapper with ado.net

xiangxiren:
related issue : #577

### DIFF
--- a/samples/Sample.RabbitMQ.MySql/Sample.RabbitMQ.MySql.csproj
+++ b/samples/Sample.RabbitMQ.MySql/Sample.RabbitMQ.MySql.csproj
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Dapper" Version="2.0.35" />
     <PackageReference Include="Pomelo.EntityFrameworkCore.MySql" Version="3.0.0-rc1.final" />
   </ItemGroup>
   <ItemGroup>
--- a/samples/Sample.RabbitMQ.SqlServer/Sample.RabbitMQ.SqlServer.csproj
+++ b/samples/Sample.RabbitMQ.SqlServer/Sample.RabbitMQ.SqlServer.csproj
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Dapper" Version="2.0.35" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="3.0.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/src/DotNetCore.CAP/Persistence/IDbConnectionExtensions.cs
+++ b/src/DotNetCore.CAP/Persistence/IDbConnectionExtensions.cs
@@ -1,0 +1,83 @@
+﻿using System;
+using System.Data;
+
+namespace DotNetCore.CAP.Persistence
+{
+    public static class IDbConnectionExtensions
+    {
+        public static int ExecuteNonQuery(this IDbConnection connection, string sql, IDbTransaction transaction = null, params object[] sqlParams)
+        {
+            if (connection.State == ConnectionState.Closed)
+            {
+                connection.Open();
+            }
+
+            using var command = connection.CreateCommand();
+            command.CommandType = CommandType.Text;
+            command.CommandText = sql;
+            foreach (var param in sqlParams)
+            {
+                command.Parameters.Add(param);
+            }
+
+            if (transaction != null)
+            {
+                command.Transaction = transaction;
+            }
+
+            return command.ExecuteNonQuery();
+        }
+
+        public static T ExecuteReader<T>(this IDbConnection connection, string sql, Func<IDataReader, T> readerFunc, params object[] sqlParams)
+        {
+            if (connection.State == ConnectionState.Closed)
+            {
+                connection.Open();
+            }
+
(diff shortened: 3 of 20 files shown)


samples/Sample.RabbitMQ.MySql/Sample.RabbitMQ.MySql.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Dapper" Version="2.0.35" />
```
yang-xiaodong, line 8: Can this reference be removed? It doesn't seem to be used

samples/Sample.RabbitMQ.SqlServer/Sample.RabbitMQ.SqlServer.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Dapper" Version="2.0.35" />
```
yang-xiaodong, line 8: Can this reference be removed? It doesn't seem to be used

test/DotNetCore.CAP.MySql.Test/DotNetCore.CAP.MySql.Test.csproj
```diff
@@ -11,6 +11,7 @@
   </ItemGroup>
   
   <ItemGroup>
+    <PackageReference Include="Dapper" Version="2.0.35" />
```
yang-xiaodong, line 14: Can this reference be removed? It doesn't seem to be used

src/DotNetCore.CAP/Persistence/IDbConnectionExtensions.cs
```diff
@@ -0,0 +1,93 @@
+﻿using System;
+using System.ComponentModel;
+using System.Data;
+
+namespace DotNetCore.CAP.Persistence
+{
+    public static class IDbConnectionExtensions
```
yang-xiaodong, line 7: The extension method is added here and marked as public. When the user refers to this library, it may cause pollution of the extension method. I suggest copying this file to put it in the storage implementation project and changing it to internal

yang-xiaodong:
It seems that the replacement the ado.net for the PostgreSql project is missing

samples/Sample.RabbitMQ.MySql/Sample.RabbitMQ.MySql.csproj
xiangxiren: image

samples/Sample.RabbitMQ.SqlServer/Sample.RabbitMQ.SqlServer.csproj
xiangxiren: image

samples/Sample.RabbitMQ.MySql/Sample.RabbitMQ.MySql.csproj
yang-xiaodong: Thank you for pointing, can we modify the example to use ado.net?

test/DotNetCore.CAP.MySql.Test/DotNetCore.CAP.MySql.Test.csproj
xiangxiren: image

samples/Sample.RabbitMQ.SqlServer/Sample.RabbitMQ.SqlServer.csproj
yang-xiaodong: Thank you for pointing, can we modify the example to use ado.net?

samples/Sample.RabbitMQ.MySql/Sample.RabbitMQ.MySql.csproj
xiangxiren: Yes, of course。However, keeping the dapper here just shows that multiple ORM frameworks can be supported。it can also add an example of ado.net

xiangxiren:
> It seems that the replacement the ado.net for the PostgreSql project is missing

oh, Because I am not familiar with the PostgreSql database, I just pulled a postgres docker image and ran a container

src/DotNetCore.CAP/Persistence/IDbConnectionExtensions.cs
xiangxiren: I have considered this issue. But if put in the implementation, there will be a copy of the same code for each project.

samples/Sample.RabbitMQ.MySql/Sample.RabbitMQ.MySql.csproj
yang-xiaodong: > Yes, of course。However, keeping the dapper here just shows that multiple ORM frameworks can be supported。it can also add an example of ado.net

For class library projects, keeping the intrusion to a minimum is more important for copying code

yang-xiaodong: > Yes, of course。However, keeping the dapper here just shows that multiple ORM frameworks can be supported。it can also add an example of ado.net

For class library projects, keeping the intrusion to a minimum is more important for copying code

src/DotNetCore.CAP/Persistence/IDbConnectionExtensions.cs
xiangxiren: ok, I will modify it

xiangxiren:
the PostgreSql project is modified

PR #588, opened 2020-06-15
Moved support for EFCore from the project DotNetCore.CAP.SqlServer to…

xiangxiren:
Hi.
This is how I implemented the porting of EFCore support to a new project.Hope to discuss with you.If you agree with this approach, I will continue to modify the two projects DotNetCore.CAP.MySql and DotNetCore.CAP.PostgreSql.


yang-xiaodong:
As for whether to implement Entityframework as a separate project, I do not agree with this at the moment. Because I think this time brings more learning costs and more nuget projects, and what we get is only one less dll file

xiangxiren:
I think.

More Nuget packages are for use on demand.For users using Ado.net or Dapper, they do not need to reference EFCore related packages.As for the learning cost, only a nuget package is added, but the way of using it has not changed.

PR #647, opened 2020-09-01
GetRuntimeMethods instead of DeclaredMethods to get inherited members.

### LINKED ISSUE #646: Topic attributes ignored due to DeclaredMethods not returning inherited methods
Environment:
- dotnet : 3.1.7
- cap version: v 3.1.0

Issue:
When using an abstract class **without** a topic attribute containing virtual methods **with** a partial topic attribute, these methods were ignored even though the derived class has a topic attribute. This was caused by using the DeclaredMethods property of the TypeInfo, changing it to GetRuntimeMethods also returns the inherited members.

Course repetition:
- create abstract class, inherit ICapSubscribe
- in this abstract class create virtual method with partial topic attribute
- create class inheriting from the abstract class with a topic attribute above the class
- check if topic is created for virtual methods

patheems:
Fixes #646


yang-xiaodong (approved):
Looks good, thanks

Issue #641, opened 2020-08-26
Custom serializer only used for interaction with message queue


pascal89:
Hi,
I implemented a custom serializer (System.Text) for interaction with the message queue (like described in https://cap.dotnetcore.xyz/user-guide/en/cap/serialization/)
All goes well when the Inbox/Outbox doesn't have to be used for a retry. When it does my custom serializer cannot serialize the deserialized json from the storage provider, because it is of type `Newtonsoft.Json.Linq.JObject` instead of the type I expected it to be.
To solve my problem I think the solution will be not to use the `DotNetCore.CAP.Serialization.StringSerializer` but the implemention of `DotNetCore.CAP.Serialization.ISerializer` for (de)serializing the message object in the storage provider.

What are your thoughts about it?

yang-xiaodong:
When restoring types from storage, we cannot know the type of data, so JObject is used here

pascal89:
The problem is the `Newtonsoft.Json` serializer does that. When using another serializer like described in https://cap.dotnetcore.xyz/user-guide/en/cap/serialization/ these kind of problems will occur when used only for sending the messages to the message queue.

Mastermindzh:
@yang-xiaodong ,

You can have a look at the migration guide from [Microsoft](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#jsondocument-and-jsonelement-compared-to-jtoken-like-jobject-jarray) to solve that issue.

We'd be happy to provide help in the form of a PR as well. Just have to know whether you wish to support more than just the Newtonsoft serializer :smile:

yang-xiaodong:
@pascal89   @Mastermindzh 

Yes, due to the reasons I described, there are still some shortcomings in custom serialization. Providing the implementation of ISerializer in the storage provider may not solve this problem I think.

Since the storage structure of the database is a serialized `Message` object, if it is in the database, we use a string for serialization. When deserializing, its type is a Message object, so it is impossible to know whether the Value field is a primitive type or a DTO object. What do you think about this?

PR is welcome, if it can solve this problem, Thanks

yang-xiaodong:
Fixed in v3.1.1

PR #664, opened 2020-09-10
Solve the issue of being restricted to using Newtonsoft

patheems:
This should solve issue #641 

It has been thoroughly tested using a MongoDB as datastore. Added some additional Tests to the Test project but didn't test it with al the other different datastores.

### DIFF
--- a/src/DotNetCore.CAP/DotNetCore.CAP.csproj
+++ b/src/DotNetCore.CAP/DotNetCore.CAP.csproj
@@ -15,6 +15,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.7.1" />
     <PackageReference Include="System.Threading.Channels" Version="4.7.1" />
+    <PackageReference Include="System.Text.Json" Version="4.7.2" />
   </ItemGroup>
  
 </Project>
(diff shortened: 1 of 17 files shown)


yang-xiaodong:
Hello @patheems , 

Thanks for your PR.   It looks good to achieve this by providing `IsJsonType` in the serialization interface, please let me test it in MySql first to make sure it works well.

Good job!

src/DotNetCore.CAP/DotNetCore.CAP.csproj
```diff
@@ -15,6 +15,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.7.1" />
     <PackageReference Include="System.Threading.Channels" Version="4.7.1" />
+    <PackageReference Include="System.Text.Json" Version="4.7.2" />
```
yang-xiaodong, line 18: `System.Text.Json` dependency are not used, Are you planning to replace `Newtonsoft.Json` with `System.Text.Json`?

yang-xiaodong (changes requested):

src/DotNetCore.CAP/DotNetCore.CAP.csproj
patheems: I just removed the reference, it was a left over that slipped in the PR. At the moment I have no plans to replace `Newtonsoft.Json` with `System.Text.Json`...could be a future change ;-)

yang-xiaodong (approved):

PR #671, opened 2020-09-17
Support leader election for need retry

recepakyuz:
If you want processing only one node for failed messages, you can use this extension.


yang-xiaodong:
Hello, why do you need this feature? In any case, consumers need to ensure idempotence

recepakyuz:
H @yang-xiaodong, I think, If the CAP run by multiple instance. Publishers can publish in per instance for same message. It can be too expensive. If we select a leader, Maybe we provide singularization.

yang-xiaodong:
> Publishers can publish in per instance for same message.

Hello,In a web application, a request reaches a specific instance through a load balancer, and the instance sends a message to the CAP. Under what circumstances will multiple instances send the same message?

recepakyuz:
I think,  If message dont process, The retry processing service process this message. But this service running multiple instance at the same time.  If this services run at the same time, they can get same message in persistent data and process same messages.  Am I wrong?

yang-xiaodong:
You are right. We have considered this scenario, we think that the error messages are a small number, and the probability of being obtained at the same time is very low. We choose to ignore this situation when consumers guarantee idempotence.

yang-xiaodong:
We already add `IsUseStorageLock` option to solve the issue in version `7.1.0-preview-193202380`.

Issue #668, opened 2020-09-15
Azure Service Bus set correlation Id and Message Id


Henery309:
Hi there,

Is there a way to set correlation id and message id for the azure service bus message? Currently we use application insight for distributed tracing and application insight can automatically group events across azure resource. But for this to happen with azure service bus we need to provide a correlation within the message. 

Message id is used to prevent duplication. Azure service bus ignore subsequent message with same message id. Currently, it looks like message id is generated by Cap library.

It would be awesome to be able to set these values from the application.

Thanks really appreciate your work on this great library.

yang-xiaodong:
Hello @Henery309 ,

Thanks for your feedback.

I think this is a feature that can be improved. We can use custom headers to achieve this. Just like in Kafka, we provide `KafkaHeader` to customize the key of the message.
https://github.com/dotnetcore/CAP/blob/1de8d6cc19c3ff4fc91bc5d17bbf27de8836d157/src/DotNetCore.CAP.Kafka/ITransport.Kafka.cs#L43-L48

We can add the conditional judgment of whether there is a custom header(CorrelationId and MessageId) in the `SendAsync` method in `AzureServiceBusTransport` and then set it to the ASB Message. 

Would you like to submit a PR to do this?

yang-xiaodong:
@Henery309 Any response?

Henery309:
Hi @yang-xiaodong, sorry been busy. More than happy to make the change. I will create a pull request by end of this week.

Henery309:
@yang-xiaodong done It looks message id was already being set via dictionary. I've added CorrelationId.

yang-xiaodong:
Thanks,  For message id I added custom set support.  Examples are as follows

```C#
var customHeader = new Dictionary<string,string>();
customHeader.Add(Headers.MessageId, "your message id");
customHeader.Add(Headers.CorrelationId, "your correlation id");
_capBus.Publish("foo", obj, customHeader);
```

yang-xiaodong:
Fixed in v3.1.1

PR #673, opened 2020-09-23
Set correlation Id for azure service bus message

Henery309:
Fix for: https://github.com/dotnetcore/CAP/issues/668


yang-xiaodong (approved):
LGTM

PR #679, opened 2020-09-29
Record the exception in the headers

xiangxiren:


src/DotNetCore.CAP/Messages/Message.cs
```diff
@@ -67,6 +67,20 @@ public static bool HasException(this Message message)
         {
             return message.Headers.ContainsKey(Headers.Exception);
         }
-    }
 
+        public static void AddOrUpdateException(this Message message, Exception ex)
+        {
+            var msg = $"{ex.GetType().Name}-->{ex}";
```
yang-xiaodong, line 73: I think we don’t need to record the exception and the stack content, it will be more troublesome to check it, only logging the exception error  message is enough

xiangxiren: Very agree

yang-xiaodong (approved):

PR #714, opened 2020-11-12
fix postgresql delete expires data logic error

cocosip:
- fix postgresql delete expires sql incorrect SQL statement


yang-xiaodong (approved):
LGTM, Thanks

PR #725, opened 2020-11-17
should not run async loop in long-running task

redochenzhen:
using async loop in long-running task is just a wast of threads in thread pool.


yang-xiaodong:
Hello, Is there any guidance or MSDN documentation for this improvement?

redochenzhen:
No. I didn't find any offical documents. Just found that the code betrays your intention.

PR #726, opened 2020-11-18
support for transactionscope

wangronghua:


yang-xiaodong:
Hi, can you describe in detail about this PR?

PR #727, opened 2020-11-22
Update nuget packages and support .Net 5

rezabayesteh:


yang-xiaodong:
Hello, thanks for your PR.

We still have some work to do in the current version, we will merge this PR in the next major version.

PR #765, opened 2021-01-13
Fixes for SNS transport implementation

roman-bezmen:
This PR contains the fixes for two issues we faced:
- If we have more than 100 topics, then AWS SDK returns a token to get the next 100 rows. This was missing.
- The first issue wasn't easy to recognize as a message is being marked as succeeded but the message actually wasn't sent in case if the topic is missing.


yang-xiaodong (approved):
LGTM, Thanks

PR #780, opened 2021-02-09
Add support for group and topic prefixes

AndriiLab:
**Problem**
We have multiple instances of the app, which deployed to AWS in different VPCs but in the same region. In this scenario, isolation works perfectly except for SNS and SQS, which are shared within the region.

**Proposed solution**
Add to `CapOptions` prefixes for topic (`TopicNamePrefix `) and group (`GroupNamePrefix `) names to make these parameters unique for each application instance. These parameters are optional, and if not defined the behavior will remain unchanged.

### DIFF
--- a/src/DotNetCore.CAP/CAP.Options.cs
+++ b/src/DotNetCore.CAP/CAP.Options.cs
@@ -35,6 +35,16 @@ namespace DotNetCore.CAP
         public string DefaultGroup { get; set; }
 
         /// <summary>
+        /// Subscriber group prefix.
+        /// </summary>
+        public string GroupNamePrefix { get; set; }
+        
+        /// <summary>
+        /// Topic prefix.
+        /// </summary>
+        public string TopicNamePrefix { get; set; }
+
+        /// <summary>
         /// The default version of the message, configured to isolate data in the same instance. The length must not exceed 20
         /// </summary>
         public string Version { get; set; }
--- a/src/DotNetCore.CAP/Internal/ConsumerExecutorDescriptor.cs
+++ b/src/DotNetCore.CAP/Internal/ConsumerExecutorDescriptor.cs
@@ -24,6 +24,8 @@ namespace DotNetCore.CAP.Internal
 
         public IList<ParameterDescriptor> Parameters { get; set; }
 
+        public string TopicNamePrefix { get; set; }
+
         private string _topicName;
         /// <summary>
         /// Topic name based on both <see cref="Attribute"/> and <see cref="ClassAttribute"/>.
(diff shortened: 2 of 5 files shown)


yang-xiaodong:
Hello,

Thanks for your PR, 

Since GroupName is customizable, so I think it is redundant to provide GroupNamePrefix  

I think it is reasonable to provide TopicNamePrefix

roman-bezmen:
@yang-xiaodong At the moment it is a bit fussy overriding the way how the group name constructs. For simple cases like providing prefixes to host multiple instances within the same infrastructure, it would be more pleasant to have a consistent way to set up prefixes for both queues and topics, even if there is an alternative way. What are your concerns?

yang-xiaodong:
What I think is that providing more configuration items will complicate the beginner and keep it simple.

Do as you say at present we have the configuration items `DefaultGroup` and `GroupNamePrefix`,  In fact, the name of `DefaultGroup` should be rename to `GroupName` so that users can know what happened.

src/DotNetCore.CAP/CAP.Options.cs
```diff
@@ -34,6 +34,16 @@ public CapOptions()
         /// </summary>
         public string DefaultGroup { get; set; }
```
yang-xiaodong, line 35: As described in the comments, rename `DefaultGroup` to `GroupName`  seemed more appropriate

AndriiLab:
Currently, we configure `GroupNamePrefix` via overriding method in `ConsumerServiceSelector.SetSubscribeAttribute`:
```C#
    public class PrefixedConsumerServiceSelector : ConsumerServiceSelector
    {
        private readonly string _prefix;

        public PrefixedConsumerServiceSelector(IServiceProvider serviceProvider, string prefix)
            : base(serviceProvider)
        {
            _prefix = prefix;
        }

        protected override void SetSubscribeAttribute(TopicAttribute attribute)
        {
            base.SetSubscribeAttribute(attribute);
            if (!string.IsNullOrEmpty(_prefix))
            {
                attribute.Group = $"{_prefix}.{attribute.Group}";
            }
        }
     }
```
And then in `Startup.cs`:
```C#
    services.AddCap(c => ...);
    services.TryRemove<IConsumerServiceSelector>();
    services.TryAddSingleton<IConsumerServiceSelector>(s => new PrefixedConsumerServiceSelector(s, "MyPerfix"));
```
Possibly, I understand this overring incorrectly, but I think there could be a more simple approach just to add prefix like
```C#
    services.AddCap(c => {
        c.GroupNamePrefix = "MyGroupPrefix",
        c.TopicNamePrefix = "MyTopicPrefix"
    });
```
Since `GroupNamePrefix ` and `TopicNamePrefix ` are not required, the application will behave as usual without these parameters. Similar behavior was implemented for suggested `DefaultGroup` (`GroupName`)

AndriiLab:
Dear @yang-xiaodong,
This feature is very important for us since we use CAP in production. If you have any concerns about releasing this feature for global configuration, I can try to implement it only for the AWS provider.

yang-xiaodong:
well, I want let you know I will rename options `DefaultGroup` to `GroupName` , thanks for your PR

yang-xiaodong:
Preview version `5.0.0-preview-131679580 ` released to nuget

AndriiLab:
Thank you very much, @yang-xiaodong!

PR #798, opened 2021-03-10
bugfix: Using with EF, transaction will not rollback when exception occurred while persistting event data

### LINKED ISSUE #796: Using with EF, transaction will not rollback when exception occurred while persistting event data
When using EntityFramework, transaction will not rollback if exception occurred while persistting event data.
I found the reason is bug in [ ICapTransaction.MySql.cs](https://github.com/dotnetcore/CAP/blob/ea1639f3622baaf6d985af111c4f208fd3839658/src/DotNetCore.CAP.MySql/ICapTransaction.MySql.cs#L87):

under EF, the transaction is `IDbContextTransaction ` not `IDbTransaction`, so to fix this, the line above should change to:
`(DbTransaction as IDisposable)?.Dispose();`

snakorse:
Closes #796


yang-xiaodong (approved):
LGTM

PR #802, opened 2021-03-18
Fix attribute in Quick Start

Martin4ndersen:
Change from `FromService` to `FromServices`.
Ref:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.fromservicesattribute


yang-xiaodong (approved):
LTGM ,  Thanks

PR #806, opened 2021-03-23
Fix topic selector in IConsumerServiceSelector

AndriiLab:
This is a fix to PR https://github.com/dotnetcore/CAP/pull/780

After additional verification, I have reverted my changes for assigning `TopicNamePrefix` to `key` in `ConsumerServiceSelector.SelectBestCandidate` because it is already contained in the key if the topic prefix was defined.

This bug affects only the case when `TopicNamePrefix` was defined in CAP options


yang-xiaodong (approved):
LGTM, Thanks

PR #815, opened 2021-03-30
Added DefaultChallengeScheme to Dashboard authentication middleware

mviegas:
This PR adds a DefaultChallengeScheme property to CAP.DashboardOptions.cs that sets up the scheme to be used when the Dashboard uses authentication via Challenge using the ASP.NET middlewares. If none is setup, the default of the application will be used.

It is useful for application with multiple authentication schemes.

Updated the sample as well.


yang-xiaodong (approved):
Looks good, Thanks

Issue #811, opened 2021-03-26
5.0.0 version dashboard page not update


luox78:
I found 5.0.0 version dashboard page not update, then i try to downgrade package to 3.1.2 this issue is fixed
![image](https://user-images.githubusercontent.com/29404988/112607253-f643cb80-8e53-11eb-9db7-f33069773b92.png)

yang-xiaodong:
Hello,

I can't reproduce your issue

Would you like to investigate this issue?

luox78:
Hello，I just run this in a test application, found the same problem

below is my test codes
![image](https://user-images.githubusercontent.com/29404988/112833350-2f38a600-90c9-11eb-94e0-02a1a07c42a9.png)
![image](https://user-images.githubusercontent.com/29404988/112833416-44153980-90c9-11eb-8a36-d6f39963a889.png)
![image](https://user-images.githubusercontent.com/29404988/112833457-57c0a000-90c9-11eb-852b-cd5cb99e628a.png)

below is runing screenshot, when i click send four times ,console print datetime.now and the sended message number should be 4,but the dashboard page donnot update
![image](https://user-images.githubusercontent.com/29404988/112833846-dddce680-90c9-11eb-9e35-97584d526c40.png)

luox78:
另外我发现dashboard打开速度特别慢，量20w左右
![image](https://user-images.githubusercontent.com/29404988/113645377-5a1e8d80-96b9-11eb-91e3-a9c8346c9fad.png)

yang-xiaodong:
列表页可以针对日期和状态列添加索引，我测试了 30 万条大概 500ms 左右，
主页慢是由于需要进行聚合运算，SQL Server 本身计算就要那么久，暂不解决

yang-xiaodong:
Issue was fixed in 5.0.1 and released to NuGet a few minutes ago

PR #819, opened 2021-04-01
Fix Dashboard metric not update #811

luox78:
This PR is to fix dashboard page not update problem，It looks like due to change the JsonSerializer, cause  js exception


yang-xiaodong (approved):
LGTM, Thanks

PR #820, opened 2021-04-02
Fix the expiration time display problem in the dashboard due to time …

xiangxiren:
1.The added and ExpiresAt assignments of the message are all local time 
2.When displayed on the dashboard, it is a timestamp calculated from the ExpiresAt and 1970-1-1

So ignore the problem of time zone


yang-xiaodong (approved):
LGTM

PR #824, opened 2021-04-05
Update faq.md

zemien:
Clarify that TableNamePrefix option is only for MySQL. Minor spelling and grammar corrections.


yang-xiaodong (approved):
It is a praisable job to do it for those of us whose English is not the first language. 

Thank you very much.

Issue #828, opened 2021-04-13
Azure Service Bus session support


jonekdahl:
Hi,

First of all, thanks for creating CAP, it seems very useful!

We're planning to use CAP to implement the outbox pattern, publishing messages via Azure Service Bus that will be consumed by non-CAP services. We would like to use [Service Bus sessions](https://docs.microsoft.com/en-us/azure/service-bus-messaging/message-sessions) to ensure that multiple messages for the same entity are delivered in FIFO order, but it seems that Service Bus sessions are not supported by CAP.

As far as we could tell, there is no way to pass a session id in a message so that [AzureServiceBusTransport](https://github.com/dotnetcore/CAP/blob/master/src/DotNetCore.CAP.AzureServiceBus/AzureServiceBusTransport.cs) will pick it up and propagate it correctly. If this is a misunderstanding, please enlighten us by showing us how to do it.

Our understanding is that, in order to implement sessions, a session id must be added as a special property in the Service Bus message. This is an example where we have extended the message construction in AzureServiceBusTransport with the `SessionId` property:

```csharp
                var message = new Microsoft.Azure.ServiceBus.Message
                {
                    MessageId = transportMessage.GetId(),
                    Body = transportMessage.Body,
                    Label = transportMessage.GetName(),
                    CorrelationId = transportMessage.GetCorrelationId(),
                    SessionId = <sessionId> // Get sessionId from somewhere, presumably message headers?
                };
 ```

Would you be interested in adding support for sessions? With your guidance, we could prepare a PR for this feature.

yang-xiaodong:
Hello Jon,

Yes, we are happy to add support for sessions. You are right, this can be read from the message header like [KafkaHeaders.cs](https://github.com/dotnetcore/CAP/blob/master/src/DotNetCore.CAP.Kafka/KafkaHeaders.cs)

PR is welcome, thanks

jonekdahl:
Hi again, there is now an initial PR for adding session support.

jonekdahl:
Thanks for merging the PR @yang-xiaodong. Would you consider making a preview release?

yang-xiaodong:
Hi @jonekdahl ,

Version `5.0.2-preview-135224594` has been released to nuget in a few minutes ago, Thanks for your PR!

thisispaulsmith:
@jonekdahl out of interest what do you use at the consumer side? Handhold or something like Masstransit?

jonekdahl:
[...]

PR #829, opened 2021-04-14
Add support for Azure Service Bus sessions

jonekdahl:
As discussed in #828, this is a PR to add session support for Azure Service Bus.

Adds a new config option `EnableSessions` for Azure Service Bus. Session support must be explicitly enabled. When disabled, CAP works like before.

When sessions are enabled:

  * `RequiresSession` is set on auto-created subscriptions
  * The `SessionId` property is set on the Service Bus Message
    * uses the value of the `cap-session-id` header, when present
    * otherwise falls back to using the message id (similar to the handling of `cap-kafka-key`)
  * RegisterSessionHandler is used instead of RegisterMessageHandler when consuming events

There is also some brief documentation on the feature, how to enable and use it (in English, possibly needs translation?).

As we don't use CAP on the consumer side ourselves, this part is unfortunately not well tested.

Suggestions welcome!


yang-xiaodong (approved):
LGTM, Thanks

Issue #834, opened 2021-04-17
Custom ConsumerServiceSelector fails to route messages with RabbitMQ


flipdoubt:
The test `ConsumerServiceSelector` implemented [here](https://github.com/dotnetcore/CAP/blob/master/test/DotNetCore.CAP.Test/CustomConsumerSubscribeTest.cs) for PR #333 fails to find a consumer for messages sent from the RabbitMQ management tool as demonstrated in my test project [here](https://github.com/flipdoubt/CAP/tree/MyConsumerServiceSelector/samples/MyConsumerSelector).

When sending a message with routing keys `Candidates.Foo`, `Candidates.Foo2`, or `Candidates.Foo3`, the console emits messages like the following:
```
warn: DotNetCore.CAP.Internal.ConsumerRegister[0]
      The Subscriber of the message(1383559940332003328) still fails after 50th executions and we will stop retrying.
```

What needs to be done to successfully route messages to `CustomInterfaceTypesClass`?

flipdoubt:
I fought my way through the issue and submitted #835 that adds [GenericConsumerServiceSelector](https://github.com/flipdoubt/CAP/blob/MyConsumerServiceSelector/samples/MyConsumerSelector/GenericConsumerServiceSelector.cs) where the subscriber interface and attribute can be supplied to the DI constainer when adding the [IConsumerServiceSelector](IConsumerServiceSelector) as shown [here](https://github.com/flipdoubt/CAP/blob/MyConsumerServiceSelector/samples/MyConsumerSelector/Program.cs#L20).

The original code for #333 fails because the consumer service selector never sets `ConsumerExecutorDescriptor`, which the PR does [here](https://github.com/flipdoubt/CAP/blob/MyConsumerServiceSelector/samples/MyConsumerSelector/GenericConsumerServiceSelector.cs#L73).

The improved [GenericConsumerServiceSelector](https://github.com/flipdoubt/CAP/blob/MyConsumerServiceSelector/samples/MyConsumerSelector/GenericConsumerServiceSelector.cs) could be useful for others.

flipdoubt:
Would it be better if I replace this in the unit test added for #333 and delete the sample I added which does not follow the sample naming convention?

yang-xiaodong:
Yes,  that will be nice!

flipdoubt:
Created PR #843.

PR #843, opened 2021-04-25
Use GenericConsumerServiceSelector in tests.

flipdoubt:
Per the last comment by @yang-xiaodong on https://github.com/dotnetcore/CAP/issues/834#issuecomment-822934344, here is the pull request that adds a unit test for subscriptions with parameters.


yang-xiaodong (approved):
LGTM, Thanks

PR #845, opened 2021-04-26
DiagnosticObserver does not use null connection

flipdoubt:
DiagnosticObserver returns rather than using a possible null SqlConnection that may come from another component.

### DIFF
--- a/src/DotNetCore.CAP.SqlServer/Diagnostics/DiagnosticObserver.cs
+++ b/src/DotNetCore.CAP.SqlServer/Diagnostics/DiagnosticObserver.cs
@@ -40,7 +40,7 @@ namespace DotNetCore.CAP.SqlServer.Diagnostics
         {
             if (evt.Key == SqlAfterCommitTransaction || evt.Key == SqlAfterCommitTransactionMicrosoft)
             {
-                var sqlConnection = (SqlConnection)GetProperty(evt.Value, "Connection");
+                if (!GetSqlConnection(evt, out SqlConnection sqlConnection)) return;
                 var transactionKey = sqlConnection.ClientConnectionId;
                 if (_bufferList.TryRemove(transactionKey, out var msgList))
                     foreach (var message in msgList)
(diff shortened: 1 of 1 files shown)


yang-xiaodong:
Do we consider removing the subscription key start with `System.Data.SqlClient.XXXXX`  so that we will not receive events from `System.Data.SqlClient` ?

flipdoubt:
I agree listening to a more focused subset of events sounds like the better way to go, but I'm not so familiar with this code nor how to test it. It appears that the subscription is setup in `DiagnosticProcessorObserver`. I searched but cannot find a different listener name using https://github.com/dotnet/SqlClient/search?q=diagnosticlistener.

src/DotNetCore.CAP.SqlServer/Diagnostics/DiagnosticObserver.cs
```diff
@@ -50,13 +50,19 @@ public void OnNext(KeyValuePair<string, object> evt)
...
             {
-                var sqlConnection = (SqlConnection)GetProperty(evt.Value, "Connection");
+                if (!GetSqlConnection(evt, out SqlConnection sqlConnection)) return;
                 var transactionKey = sqlConnection.ClientConnectionId;
 
                 _bufferList.TryRemove(transactionKey, out _);
             }
         }
 
+        private bool GetSqlConnection(KeyValuePair<string, object> evt, out SqlConnection sqlConnection)
```
yang-xiaodong, line 60: I think to use `TryGetSqlConnection` as the method name and  defined as the static method will make more sense

yang-xiaodong:
@flipdoubt  Do you know what scenarios the Connection property will be null? It sounds like an impossible thing in the transaction commit process

src/DotNetCore.CAP.SqlServer/Diagnostics/DiagnosticObserver.cs
flipdoubt: Yes, `TryGetSqlConnection` fits better with Microsoft APIs. I pushed this change as well. Thanks.

yang-xiaodong (approved):
LGTM

flipdoubt:
> @flipdoubt Do you know what scenarios the Connection property will be null?

@yang-xiaodong, you are correct that the underlying connection is never null, but using `as Microsoft.Data.SqlClient.SqlConnection` yields null when the underlying connection is a `System.Data.SqlClient.SqlConnection`, such as when Hangfire creates the connection. The original `(SqlConnection) GetProperty(...)` throws `InvalidCastException` because Hangfire uses `System.Data.SqlClient.SqlConnection` instead of `Microsoft.Data.SqlClient.SqlConnection`. Guarding against null is just defensive programming where other components still use `System.Data.SqlClient.SqlConnection`.

Another work around would be to duplicate this code for `System.Data.SqlClient.SqlConnection`, but that requires referencing the "system" package.

PR #846, opened 2021-04-26
Fix null reference in AmazonSQSTransport

AndriiLab:
**Problem**
The `NullReferenceException` by `DotNetCore.CAP.Messages.MessageExtensions.AddOrUpdateException` when no SNS topics found in `AmazonSQSTransport`.

**Proposed solution**
This is related to the usage old overload of `OperateResult` in `AmazonSQSTransport`, which does not accept an exception. It is not valid anymore, since `MessageSender` expecting that exception is not null when message failed.
Removed obsolete overload `OperateResult Failed(params OperateError[] errors)` and used a recent one in `AmazonSQSTransport`


yang-xiaodong (approved):
LGTM, Thanks

Issue #861, opened 2021-05-09
RabbitMQ Connection close forced


hetaoos:
Hi, After upgrading from 5.0.0 to 5.0.1 or 5.0.2, cap cannot connect to the RabbitMQ server and reports "Connection close forced" exception. 

When downgrade to 5.0.0, the exception does not occur. 

RabbitMQ 3.8.14 Erlang 23.3.1

log:
```
2021-05-09T02:56:33.151799434Z fail: DotNetCore.CAP.Internal.ConsumerRegister[0]
2021-05-09T02:56:33.151825096Z       The AMQP operation was interrupted: AMQP close-reason, initiated by Application, code=200, text='Connection close forced', classId=0, methodId=0
2021-05-09T02:56:33.151833874Z       RabbitMQ.Client.Exceptions.OperationInterruptedException: The AMQP operation was interrupted: AMQP close-reason, initiated by Application, code=200, text='Connection close forced', classId=0, methodId=0
2021-05-09T02:56:33.151843336Z          at RabbitMQ.Client.Impl.SimpleBlockingRpcContinuation.GetReply(TimeSpan timeout)
2021-05-09T02:56:33.151850829Z          at RabbitMQ.Client.Impl.ModelBase.ModelRpc(MethodBase method, ContentHeaderBase header, Byte[] body)
2021-05-09T02:56:33.151858402Z          at RabbitMQ.Client.Framing.Impl.Model._Private_QueueBind(String queue, String exchange, String routingKey, Boolean nowait, IDictionary`2 arguments)
2021-05-09T02:56:33.151866082Z          at DotNetCore.CAP.RabbitMQ.RabbitMQConsumerClient.Subscribe(IEnumerable`1 topics)
2021-05-09T02:56:33.151873324Z          at DotNetCore.CAP.Internal.ConsumerRegister.<>c__DisplayClass17_1.<Start>b__2()
2021-05-09T02:56:33.158342427Z warn: DotNetCore.CAP.RabbitMQ.ConnectionChannelPool[0]
2021-05-09T02:56:33.158368148Z       RabbitMQ client connection closed! --> Connection close forced
2021-05-09T02:56:33.164347969Z warn: DotNetCore.CAP.RabbitMQ.ConnectionChannelPool[0]
2021-05-09T02:56:33.164397986Z       RabbitMQ client connection closed! --> Connection close forced
[...]

PR #863, opened 2021-05-10
Fix RabbitMQ connection was closed unexpectedly.

hetaoos:
In the `Dispose` method of `RabbitMQConsumerClient`, the connection should not be closed because the connection is reused.  

#861


yang-xiaodong (approved):
Looks good , Thanks

PR #868, opened 2021-05-12
Implement IInfrastructure` on CapEFDbTransaction

### LINKED ISSUE #867: ICapTransaction on CapEFDbTransaction is private.
It's currently not possible to access `ICapTransaction` from `CapEFDbTransaction` which makes Dapper integration difficult as you can't access the underlying transaction. 

Any objection to a PR to make this public?

thisispaulsmith:
Resolves #867. Enables access to the underlying transaction when using Entity Framework.


yang-xiaodong (approved):
Agree,

Implementing `IInfrastructure<DbTransaction>` helps to call the `GetDbTransaction` extension method through `IDbContextTransaction`.

Thank you

PR #869, opened 2021-05-13
make OperateResult._errors readonly

WeihanLi:
Since there's no set opertions on `_errors`, we could make it `readonly`


yang-xiaodong (approved):

yang-xiaodong:
Thanks

PR #882, opened 2021-05-20
Fix bug: Dashboard not show message list

alexpee:
Fix the bug that the dashboard (useSQLServer) doesn't show the message list, with error : the sqlparameter is already contained by another sqlparametercollection 
The cause is due to the reusing sqlparam in SqlServerMonitoringApi.cs,  or the DbConnectionExtensions helper doesn't clear up command.param in every executescalar and executereader call


xiangxiren:
#881

alexpee:
thanks for update. I will close the PR

PR #891, opened 2021-05-25
Implementing IInfrastructure interface in order to the GetDbTransaction extension method of EF core work

manelferreira:
Applying the same fix realized for sql server transaction interface ([issue explained here](https://github.com/dotnetcore/CAP/issues/867))  to the postgresql transaction interface.

Brief overview: the problem is that the EF Core GetDbTransaction method ([extension method reference](https://github.com/dotnet/efcore/blob/main/src/EFCore.Relational/Storage/DbContextTransactionExtensions.cs)) is throwing InvalidOperationException because the CAP PostgreSQL transaction class is not implementing the required interface so the extension method can be executed.

### DIFF
--- a/src/DotNetCore.CAP.PostgreSql/IDbContextTransaction.CAP.cs
+++ b/src/DotNetCore.CAP.PostgreSql/IDbContextTransaction.CAP.cs
@@ -2,14 +2,16 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Data.Common;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


wagnergraciano:
I have the same issue

VictorSega:
Cool!
I'm with the same problem, thanks for the fix!
I hope this is merged quickly.

xecoo:
How did you manage to arrive at this solution?? Because recently I aborted my project because I didn't find any solution until today with this PR.

yang-xiaodong (approved):
LGTM , Thanks

PR #900, opened 2021-06-02
Add try catch on Dispatcher.Processing when invoke subscriber

luox78:


yang-xiaodong:
Hello, There is a Try Catch inside the function to handle exceptions, there is no need to add additional

luox78:
> Hello, There is a Try Catch inside the function to handle exceptions, there is no need to add additional

你好，里面catch处理的时候有个设置消息失败状态的操作，很可能有数据库导致抛出异常

yang-xiaodong (approved):
LGTM

PR #911, opened 2021-06-20
Abstract base class for ISubscribeFilter

teomane:
An abstract base class implementation will be useful for ISubscribeFilter like as ISaveChangesInterceptor and SaveChangesInterceptor in entity framework core.


yang-xiaodong:
Hello  @teomane , 

We currently do not support the use of multiple filters. Do you think it is more reasonable to use abstract class?

teomane:
Hi ! In the case of the abstract class, I have an option to use just the methods I need. In my case, I will use the only OnSubscribeExecuting method. Otherwise, I have to implement other methods with a blank body.

yang-xiaodong (approved):

Issue #907, opened 2021-06-15
[Feature request] Support CancellationTokens on Subscriber handlers


matt-psaltis:
Currently we can pass through CapHeader as a Handle method parameter. CAP picks this up as IsFromCap in the ParameterDescriptor. Can this be extended to CancellationTokens for Async methods? This cancellation token should register with the IBootstrapper.BootstrapAsync cancellation token so that shutdown of the Bootstrapper initiates cancellation of Subscribers.

Example:

```
[Topic("Workshop", typeof(CompanyCreated))]
public Task Handle(CompanyCreated message, CancellationToken cancellationToken) {
}
```

Happy to help with these requests if you're ok with the approach of course :D

yang-xiaodong:
Hello,

This  suggestion looks good, would you like to submit a PR?

yang-xiaodong:
Fixed in `5.1.1-preview-141622241`

Issue #908, opened 2021-06-15
[Bug] The Method Executor Cache does not distinguish between different classes when the Generic base class defines the subscription


matt-psaltis:
Given the following class structure, the cache key does not factor in open generic typed classes.

At runtime, we use typeof(BaseClass<>)
   .MakeGenericType(typeof(MessageType1..n)) to generate closed type classes.

In .NET Methods declared in the open generic BaseClass return the same ReflectedType name and method MetadataToken even when the method is retrieved via the closed generic type subclass.

Because of this, MessageType1 is actually processed by:
`public void Handle(MessageType1 message)`
`public void Handle(MessageType2 message)`

A cast exception is thrown for the second method.

The change I would like to make is at the following location:
https://github.com/dotnetcore/CAP/blob/55cc2c0a3fefca4c171233b6c67d3441035fa314/src/DotNetCore.CAP/Internal/ISubscribeInvoker.Default.cs#L43

The test below shows that if the cache key is changed to include the method.ReflectedType.TypeHandle.Value we avoid this key collision. We might also save a little bit of memory in some cases where the type name is longer than the handle :D

```
        [Fact]
        public void Test()
        {
            var method1 = typeof(BaseClass<>)
                          .MakeGenericType(typeof(MessageType1))
                          .GetMethod(nameof(BaseClass<object>.Handle));
            var method2 = typeof(BaseClass<>)
                          .MakeGenericType(typeof(MessageType2))
                          .GetMethod(nameof(BaseClass<object>.Handle));

            Assert.Equal(method1.MetadataToken, method2.MetadataToken);
            Assert.NotEqual(method1.ReflectedType.TypeHandle.Value, method2.ReflectedType.TypeHandle.Value);
        }

    public class MessageType1 {}
    public class MessageType2 {}

    public abstract class BaseClass<T>
    {
        public void Handle()
        {

        }
    }

```

yang-xiaodong:
Hello, 

For #666, #695  PRs related to cache key changes, we need to evaluate the feasibility of using `method.ReflectedType.TypeHandle.Value`

matt-psaltis:
Is there some initial feasibility you would like me to contribute to? Tests etc?

yang-xiaodong:
Yes, I mean to do some testing. I just tested it and it seems works good. Would you like to submit a PR?

matt-psaltis:
Awesome, happy to.

matt-psaltis:
[...]

PR #912, opened 2021-06-21
Add cancellation token support to Subscribers

matt-psaltis:
Relates to: https://github.com/dotnetcore/CAP/issues/907

Wanted to get your thoughts on this currently. I've added an example integration test as well as some additional unit tests for the subscriber resolution when cancellation tokens are in play. There's a couple of signature changes that would be potentially breaking so I'd like to get your thoughts on those please?

### DIFF
--- a/src/DotNetCore.CAP/Internal/ICapPublisher.Default.cs
+++ b/src/DotNetCore.CAP/Internal/ICapPublisher.Default.cs
@@ -38,11 +38,6 @@ namespace DotNetCore.CAP.Internal
 
         public AsyncLocal<ICapTransaction> Transaction { get; }
 
-        public void Start(CancellationToken stoppingToken)
-        {
-            _dispatcher.Start(stoppingToken);
-        }
-
         public Task PublishAsync<T>(string name, T value, IDictionary<string, string> headers, CancellationToken cancellationToken = default)
         {
             return Task.Run(() => Publish(name, value, headers), cancellationToken);
--- a/src/DotNetCore.CAP/Internal/ICapPublisher.Default.cs
+++ b/src/DotNetCore.CAP/Internal/ICapPublisher.Default.cs
@@ -38,6 +38,11 @@ namespace DotNetCore.CAP.Internal
 
         public AsyncLocal<ICapTransaction> Transaction { get; }
 
+        public void Start(CancellationToken stoppingToken)
+        {
+            _dispatcher.Start(stoppingToken);
+        }
+
         public Task PublishAsync<T>(string name, T value, IDictionary<string, string> headers, CancellationToken cancellationToken = default)
         {
             return Task.Run(() => Publish(name, value, headers), cancellationToken);
--- a/src/DotNetCore.CAP/Internal/ISubscribeInvoker.Default.cs
+++ b/src/DotNetCore.CAP/Internal/ISubscribeInvoker.Default.cs
@@ -36,7 +36,7 @@ namespace DotNetCore.CAP.Internal
             cancellationToken.ThrowIfCancellationRequested();
 
             var methodInfo = context.ConsumerDescriptor.MethodInfo;
-            var reflectedType = methodInfo.ReflectedType.Name;
+            var reflectedType = methodInfo.ReflectedType.TypeHandle.Value;
 
             _logger.LogDebug("Executing subscriber method : {0}", methodInfo.Name);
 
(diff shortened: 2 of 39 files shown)


src/DotNetCore.CAP/Internal/ICapPublisher.Default.cs
```diff
@@ -38,6 +38,11 @@ public CapPublisher(IServiceProvider service)
 
         public AsyncLocal<ICapTransaction> Transaction { get; }
 
+        public void Start(CancellationToken stoppingToken)
```
yang-xiaodong, line 41: Should it be removed?  I think the prevent Start method in CapPublisher looks a bit strange?  and the method is not called

src/DotNetCore.CAP/Internal/ISubscribeInvoker.Default.cs
```diff
@@ -36,7 +36,7 @@ public async Task<ConsumerExecutedResult> InvokeAsync(ConsumerContext context, C
             cancellationToken.ThrowIfCancellationRequested();
 
             var methodInfo = context.ConsumerDescriptor.MethodInfo;
-            var reflectedType = methodInfo.ReflectedType.Name;
+            var reflectedType = methodInfo.ReflectedType.TypeHandle.Value;
```
yang-xiaodong, line 39: If we use `ReflectedType.TypeHandle.Value`, as described in #908, do we still need `"{methodInfo.Module.Name}_{reflectedType}_{methodInfo.MetadataToken}"` to ensure that the key is unique?

yang-xiaodong (commented):
Hello @matt-psaltis,

Thank you very much for your PR,  I reviewed and found two issues, please confirm it

yang-xiaodong:
Hello @matt-psaltis , Any further response on this PR?

src/DotNetCore.CAP/Internal/ISubscribeInvoker.Default.cs
matt-psaltis: Hey got slammed with work this week. I was trying to keep the PRs separate so I was planning to do this once I had some test coverage in place in a subsequent PR. happy to amalgamate tho if that's easier.

src/DotNetCore.CAP/Internal/ICapPublisher.Default.cs
matt-psaltis: It does look strange on the publisher side agreed. I need to come back to this and reassess. It didn't quite fit the current structure.

matt-psaltis:
> Hello @matt-psaltis , Any further response on this PR?

@yang-xiaodong I've pushed some updates and additional tests for the cache key problem from #908 Would love your thoughts when you can please.

yang-xiaodong:
Hello @matt-psaltis ,

Since the object that implements `IProcessingServer` will be started in bootstrap so what do you think  about making the `IDispatcher`  inherit the `IProcessingServer` ? 

https://github.com/dotnetcore/CAP/blob/master/src/DotNetCore.CAP/Internal/IBootstrapper.Default.cs#L33

matt-psaltis:
> Hello @matt-psaltis ,
> 
> Since the object that implements `IProcessingServer` will be started in `Bootstrapper` class so what do you think about making the `IDispatcher` inherit the `IProcessingServer` ?
> 
> https://github.com/dotnetcore/CAP/blob/master/src/DotNetCore.CAP/CAP.ServiceCollectionExtensions.cs#L47
> https://github.com/dotnetcore/CAP/blob/master/src/DotNetCore.CAP/Internal/IBootstrapper.Default.cs#L33

Awesome! I actually did go down that path today. I removed it because I wasn't sure on the intention of a "processing server" as a concept vs the dispatcher. I'll consolidate, it was cleaner that way, thanks.

matt-psaltis:
> Hello @matt-psaltis ,
> 
> Since the object that implements `IProcessingServer` will be started in `Bootstrapper` class so what do you think about making the `IDispatcher` inherit the `IProcessingServer` ?
> 
> https://github.com/dotnetcore/CAP/blob/master/src/DotNetCore.CAP/CAP.ServiceCollectionExtensions.cs#L47
> https://github.com/dotnetcore/CAP/blob/master/src/DotNetCore.CAP/Internal/IBootstrapper.Default.cs#L33

I remember why I unwound this now, it was because of the number of places that directly reference the IDispatcher.

I could register it as an IProcessingServer using something like this:
```
            services.TryAddEnumerable(ServiceDescriptor.Singleton<IProcessingServer, IDispatcher>(sp =>
                sp.GetRequiredService<IDispatcher>()));
```

Need to make sure we only get a single instance created so by registering it as an IProcessing Server as well as IDispatcher, we need to use a factory pattern like above to reference the same IDispatcher instance. Does that make sense?
Happy to do it this way - I'm just being overly cautious of making drastic changes like I would with my own repositories :D

yang-xiaodong:
It’s a good idea to use a factory to resolve, just do it

matt-psaltis:
> It’s a good idea to use a factory to resolve, just do it

Done.

yang-xiaodong (approved):

yang-xiaodong:
Looks good, thanks very much

PR #921, opened 2021-06-26
fix:  health check path is changed to 'cap/api/health'

wwwu:
1，修正health check 路径为 'cap/api/health'
2，开放CapCache，可自实现INodeDiscoveryProvider


yang-xiaodong (approved):
LGTM, Thanks

PR #924, opened 2021-06-30
fix: 修复分页查询

ALer-R:
1.增加汇总接口
2.修复分页查询


yang-xiaodong:
Hello,

Thanks for your PR,  the branch `supports/gcp-pubsub` is discarded.

Your PR seems to be a fix for the dashboard paging bug of CAP version before 5.1.0.  We're no longer provide support for the old version.

ALer-R:
项目已稳定，不可随意迁移

PR #976, opened 2021-08-19
add opentelemetry tracing to rabbitmq

meysamhadeli:
Hi.
I added opentelemetry tracing to rabbitmq and it can be use with jaeger or zipkin and ...

### DIFF
--- a/src/DotNetCore.CAP.RabbitMQ/ITransport.RabbitMQ.cs
+++ b/src/DotNetCore.CAP.RabbitMQ/ITransport.RabbitMQ.cs
@@ -18,100 +18,101 @@ using RabbitMQ.Client;
 namespace DotNetCore.CAP.RabbitMQ
 {
      internal sealed class RabbitMQTransport : ITransport
+    {
+        private readonly IConnectionChannelPool _connectionChannelPool;
+        private readonly ILogger _logger;
+        private static readonly ActivitySource ActivitySource = new ActivitySource(nameof(ITransport));
+        private static readonly TextMapPropagator Propagator = Propagators.DefaultTextMapPropagator;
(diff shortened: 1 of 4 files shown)


src/DotNetCore.CAP.RabbitMQ/ITransport.RabbitMQ.cs
```diff
@@ -2,74 +2,116 @@
...
+                    if (channel != null)
+                    {
+                        _connectionChannelPool.Return(channel);
+                    }
                 }
             }
+            
+            private void AddMessagingTags(Activity activity, TransportMessage message)
+            {
+                activity?.SetTag("message", Encoding.UTF8.GetString(message.Body.ToArray()));
```
mviegas, line 103: I see where you coming from here, but should the message really be logged?! Imagine a scenario where a message contains PII and you cannot log it due to GDPR concerns.

meysamhadeli: @mviegas Yes, It's true, I removed this line and add correlation_id and message_id instead of it.

yang-xiaodong:
Thank you for your PR,

In fact, we should create a new bridge project for opentelemetry instead of adding it to the current transport, and use the existing support features for [diagnostics events](https://cap.dotnetcore.xyz/user-guide/en/monitoring/diagnostics/#diagnostics-events) to adapt opentelemetry.

CAP’s support extension library for Skywalking has a good example that demonstrates this
https://github.com/SkyAPM/SkyAPM-dotnet/tree/master/src/SkyApm.Diagnostics.CAP

yang-xiaodong:
Closed by #1061

PR #995, opened 2021-09-06
Add use TopicName as cap-msg-name when not provided

pmdevers:
If the cap-msg-name is not provided in the headers use the topicName.

Because we cannot guarantee that the cap-msg-name is always provided. because of other producers not using Cap.

### DIFF
--- a/src/DotNetCore.CAP.Kafka/KafkaConsumerClient.cs
+++ b/src/DotNetCore.CAP.Kafka/KafkaConsumerClient.cs
@@ -96,6 +96,12 @@ namespace DotNetCore.CAP.Kafka
                     var val = header.GetValueBytes();
                     headers.Add(header.Key, val != null ? Encoding.UTF8.GetString(val) : null);
                 }
+
+                if (!headers.ContainsKey(Messages.Headers.MessageName))
+                {
+                    headers.Add(Messages.Headers.MessageName, consumerResult.Topic);
+                }
+
                 headers.Add(Messages.Headers.Group, _groupId);
 
                 if (_kafkaOptions.CustomHeaders != null)


yang-xiaodong:
Thank you for your PR, it is currently possible to do this by customizing the Header, see https://cap.dotnetcore.xyz/user-guide/en/transport/kafka/#customheaders-options

pmdevers:
cap-msg-name is not a custom header and it is always required. It throws an exception if it is not provided when resolving the coresponding provider.

And the execption did not happen with an earlier version probably 3.x

yang-xiaodong:
> cap-msg-name is not a custom header and it is always required. It throws an exception if it is not provided when resolving the coresponding subscriber.

Yes, If the message is not published by CAP, then an exception will be thrown, which we mentioned in the document.

One of our purposes for providing the CustomHeader option is to deal with problems between heterogeneous systems or non-CAP messages.

pmdevers:
I would expect some default behaviour or beter exception handling in this case it throw an unexplaind exception which is unclear what is the cause.

pmdevers:
`[16:24:40 ERR] An exception occurred when process received message. Message:'DotNetCore.CAP.Messages.TransportMessage'.
System.ArgumentNullException: Value cannot be null. (Parameter 'input')
   at System.Text.RegularExpressions.ThrowHelper.ThrowArgumentNullException(ExceptionArgument arg)
   at System.Text.RegularExpressions.Regex.IsMatch(String input)
   at DotNetCore.CAP.Internal.ConsumerServiceSelector.MatchWildcardUsingRegex(String key, IReadOnlyList`1 executeDescriptor)
   at DotNetCore.CAP.Internal.MethodMatcherCache.TryGetTopicExecutor(String topicName, String groupName, ConsumerExecutorDescriptor& matchTopic)
   at DotNetCore.CAP.Internal.ConsumerRegister.<>c__DisplayClass22_0.<RegisterMessageProcessor>b__0(Object sender, TransportMessage transportMessage)
`

rinckd:
Any other thoughts about this?

Currently, if a message does not have 'cap-msg-name', the IConsumerRegister.Default.cs will throw `Value cannot be null. (Parameter 'key')`. This will lock the consumer up at this message. 

In larger projects, it is very common for someone to write a message to a topic and get this wrong. It makes it harder to communicate to everyone that this requirement is mandatory without using a Schema Registry.

Adding the cap-msg-name in the Custom Headers is non-ideal, since that would give you a `System.ArgumentException: An item with the same key has already been added. Key: cap-msg-name`

```
        options.CustomHeaders = kafkaResult => new List<KeyValuePair<string, string>>
        {
            new ("cap-msg-name", kafkaResult.Topic.ToString())
        };
```

I feel that either the messages that don't have a 'cap-msg-name' should be explicitly ignored and Consumed (Committed), or the cap-msg-name should default to topic if null.

yang-xiaodong:
Closed by https://cap.dotnetcore.xyz/user-guide/en/cap/messaging/#custom-headers

PR #1018, opened 2021-10-18
Do not update message Content when change state.

lhysrc:
If the subscription model is a reference type, the properties of the updated model will be saved to the database.
I don't think it is necessary. The important thing is that when RETRY, the content is not the original value.


yang-xiaodong:
We need record the exception in the headers, see https://github.com/dotnetcore/CAP/pull/679

lhysrc:
Can we make a model copy before pass to subscribe method?
Otherwise, when there is an exception, if code related to the model property when retrying, the execution result will be inconsistent.

PR #1034, opened 2021-11-03
Added possibility to process messages for each consumer group independently

dario-l:
This addition do not change anything to default behavior of CAP.

To run processing messages for each consumer group independently must set option `UseDispatchingPerGroup` to `true`. This will add to DI container `DispatcherPerGroup` instead of default `Dispatcher`.

With default `Dispatcher` behavior of CAP is:

dispatcher default

As you can see `VeryFastProcessingReceiver` waits until `XSlowProcessingReceiver` finishes.

With `DispatcherPerGroup` behavior of CAP is:

dispatcher pergroup

Here we have situation where all subscribers pushes received messages to own dispatching channel. Each channel has set thread count to `ConsumerThreadCount` value.

New feature is used in new sample application `Sample.RabbitMQ.SqlServer.DispatcherPerGroup`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,8 @@ public void ShowTime2(DateTime datetime)
 }
 
 ```
-`ShowTime1` and `ShowTime2` will be called at the same time.
+`ShowTime1` and `ShowTime2` will be called one after another because all received messages are processed linear.
+You can change that behaviour increasing `ConsumerThreadCount`.
 
 BTW, You can specify the default group name in the configuration:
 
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -227,7 +227,8 @@ public void ShowTime2(DateTime datetime)
 }
 
 ```
-`ShowTime1` and `ShowTime2` will be called at the same time.
```
dario-l, line 230: That is not true if `ConsumerThreadCount=1`

yang-xiaodong (approved):
LGTM , Thanks

dario-l:
Great, thanks 😃 👍🏻

PR #1037, opened 2021-11-06
Update OnConsumerReceived event for headers

ustaserdar:
It prevents throwing unhandled exceptions about explicit casting header values to byte array

### DIFF
--- a/src/DotNetCore.CAP.RabbitMQ/RabbitMQConsumerClient.cs
+++ b/src/DotNetCore.CAP.RabbitMQ/RabbitMQConsumerClient.cs
@@ -172,6 +172,25 @@ namespace DotNetCore.CAP.RabbitMQ
             };
             OnLog?.Invoke(sender, args);
         }
+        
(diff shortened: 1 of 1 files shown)


src/DotNetCore.CAP.RabbitMQ/RabbitMQConsumerClient.cs
```diff
@@ -172,6 +172,25 @@ private void OnConsumerRegistered(object sender, ConsumerEventArgs e)
             };
             OnLog?.Invoke(sender, args);
         }
+        
+        private byte[] ObjectToByteArray(object obj)
```
yang-xiaodong, line 176: I don’t think this is a good solution, it will cause extra performance overhead for users who don’t use this feature.

yang-xiaodong:
Closed, fixed in https://github.com/dotnetcore/CAP/commit/f1500fd38228678ab5050ee8474d1f71822bce86

PR #1052, opened 2021-12-01
Update SnowflakeId algorithm

handloong:


yang-xiaodong:
https://github.com/hueifeng/snowflake-csharp/blob/master/Snowflake/SnowFlake.cs#L69

PR #1065, opened 2022-01-07
Optimize SnowFlake algorithm.

Allen-dududu:
The idea of Optimize SnowFlate algorithm comes from Seata which is an open-source distributed transaction solution of Ali. It also uses snowflake to generate Id, but they upgraded the algorithm to avoid time step back exception.

FYI: https://seata.io/zh-cn/blog/seata-analysis-UUID-generator.html

### DIFF
--- a/src/DotNetCore.CAP/Internal/SnowflakeId.cs
+++ b/src/DotNetCore.CAP/Internal/SnowflakeId.cs
@@ -2,48 +2,75 @@
 // An object that generates IDs. This is broken into a separate class in case we ever want to support multiple worker threads per process
 
 using System;
+using System.Linq;
+using System.Net.NetworkInformation;
+using System.Threading;
 
 namespace DotNetCore.CAP.Internal
 {
     public class SnowflakeId
     {
+        /// <summary>
+        /// Start time 2010-11-04 09:42:54
+        /// </summary>
         public const long Twepoch = 1288834974657L;
 
-        private const int WorkerIdBits = 5;
-        private const int DatacenterIdBits = 5;
-        private const int SequenceBits = 12;
-        private const long MaxWorkerId = -1L ^ (-1L << WorkerIdBits);
-        private const long MaxDatacenterId = -1L ^ (-1L << DatacenterIdBits);
+        /// <summary>
+        /// The number of bits occupied by workerId
+        /// </summary>
+        private const int WorkerIdBits = 10;
 
-        private const int WorkerIdShift = SequenceBits;
-        private const int DatacenterIdShift = SequenceBits + WorkerIdBits;
-        public const int TimestampLeftShift = SequenceBits + WorkerIdBits + DatacenterIdBits;
-        private const long SequenceMask = -1L ^ (-1L << SequenceBits);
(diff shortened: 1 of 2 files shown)


Allen-dududu:
Unbind time from the operating system timestamp. The generator only gets the current system timestamp at initialization as the initial timestamp, but is no longer in sync with the system timestamp.  Its subsequent increment is driven only by the increment of the serial number.  For example, if the current sequence number is 4095 and the next request comes in, the sequence number +1 overflows the 12-bit space, the sequence number is reset to zero, and the overflow carry is added to the timestamp, making the timestamp +1

yang-xiaodong:
Have you tested this algorithm?

Why is the value of recreated instance always begin with 4989988387126509569?

Allen-dududu:
> 

Sorry for that. I forgot to init the first timestamp and sequence, Fixed now.

src/DotNetCore.CAP/Internal/SnowflakeId.cs
```diff
@@ -59,61 +86,96 @@ public static SnowflakeId Default()
...
-            while (timestamp <= lastTimestamp) timestamp = TimeGen();
-            return timestamp;
+            NetworkInterface[] nics = NetworkInterface.GetAllNetworkInterfaces();
+
+            if (nics == null || nics.Length < 1)
+            {
+                throw new Exception("no available mac found");
+            }
+
+            var adapter = nics.FirstOrDefault();
```
yang-xiaodong, line 176: Better to `var adapter = nics[0]`

src/DotNetCore.CAP/Internal/SnowflakeId.cs
```diff
@@ -59,61 +86,96 @@ public static SnowflakeId Default()
...
+        /// auto generate workerId, try using mac first, if failed, then randomly generate one
+        /// </summary>
+        /// <returns>workerId</returns>
+        public static long GenerateWorkerId(int maxWorkerId)
+        {
+            try
+            {
+                return GenerateWorkerIdBaseOnMac();
+            }
+            catch (Exception e)
```
yang-xiaodong, line 157: Exception e ,  e is not used here

src/DotNetCore.CAP/Internal/SnowflakeId.cs
```diff
@@ -59,61 +86,96 @@ public static SnowflakeId Default()
...
+        /// </summary>
+        /// <returns></returns>
+        private long GetNewestTimestamp()
+        {
+            return DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - Twepoch;
+        }
+    }
 
-                return id;
+    static class Util
```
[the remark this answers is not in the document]
Allen-dududu: Actually, It is internal.  https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1527?f1url=%3FappId%3Droslyn%26k%3Dk(CS1527)

Ok, For clear, I will use Explicit Declaration. "internal static class Util"

yang-xiaodong (approved):
LGTM, Thanks

PR #987, opened 2021-08-24
Replace Dictionary with ConcurrentDictionary

xiangxiren:

PR #1104, opened 2022-03-22
used concurrentdictionary since PublishedMessages and ReceivedMessage…

wakiter:
used a concurrentdictionary since PublishedMessages and ReceivedMessages are public and accessed from various places

if one thread enumerates through the dictionary and the other tries to update it, we get a NullReferenceException


yang-xiaodong:
see #987

yang-xiaodong (approved):
LGTM

PR #1125, opened 2022-04-15
Add Exception Details

eymentopcuoglu:
Hello,

Currently, `cap-exception` header does not include the exception stack trace, which would be useful. So, I suggest adding another header called `cap-exception-details` which holds `exception.ToString()`.


yang-xiaodong:
Hello, thanks your PR.
Although the exception stack detail is useful, but it is not in line with our design, we want users to see the exception stack of the error in the logs, not in the content field of the data table.

eymentopcuoglu:
For anybody landing here, wishing to customize this behavior: 
Adding this header is also possible via [filters](https://cap.dotnetcore.xyz/user-guide/en/cap/filter/). Here is the snippet:

```cs
public void OnSubscribeException(ExceptionContext context)
{
    context.DeliverMessage.Headers["cap-exception-details"] = context.Exception.ToString();
}
```

Issue #1148, opened 2022-05-31
Redis Streams deployed in IIS, after the application pool is recycled an exception occurs occasionally 


cdefg198:
Hello,
I have a problem,

1. Production Environment:   IIS10、CAP 6.0.1、CAP.RedisStreams 6.0.1、CAP.SqlServer 6.0.1
2. the consumer website application pool is recycled and self started at 5 a.m. every day
3. Exception msg:
`DotNetCore.CAP.RedisStreams.AsyncLazyRedisConnection：Server replied with error, BUSYGROUP Consumer Group name already exists, for endpoint`
`DotNetCore.CAP.Internal.ConsumerRegister：BUSYGROUP Consumer Group name already exists`

4.Scene:   The CAP Consumer Website is ok, cap.Published table exists record,also StatusName = Succeeded, but cap.Received no record,restart the  CAP Consumer Website after finding problems,the system is ok.

cap.Published
![1](https://user-images.githubusercontent.com/3070114/171127751-840b7b15-9099-44a3-ade6-b70bd8e6e6b3.jpg)

cap.Received
![2](https://user-images.githubusercontent.com/3070114/171127852-a5389e0a-c952-4af0-b540-3ec5d706206e.jpg)

cap.published  2022-05-31 08:12:17-2022-05-31 09:11:48 interval data no write to cap.received

windwos event error msg
![3](https://user-images.githubusercontent.com/3070114/171128929-68e63160-09d9-424b-b691-eb583ff27587.jpg)

![4](https://user-images.githubusercontent.com/3070114/171128964-98a84f8a-c706-4471-992a-39adb74a20a5.jpg)

![5](https://user-images.githubusercontent.com/3070114/171128978-b3240875-dac5-4a16-a720-a8588a95d3f6.jpg)

MahmoudSamir101:
Hi @cdefg198 , Is this limited to iis pool recycling? This behavior can also be reproduced by running the consumer outside of iis and restarting it.

I will investigate more from my side to run my tests againt iis

cdefg198:
> Hi @cdefg198 , Is this limited to iis pool recycling? This behavior can also be reproduced by running the consumer outside of iis and restarting it.
> 
> I will investigate more from my side to run my tests againt iis

My redis version is **window redis-x64-5.0.10**,after deployment to the production environment, there are two occurrences in total.


The following two exceptions occur occasionally in the test environment, but do not affect the use，
[...]

PR #1150, opened 2022-06-03
Change reading/creating streams and consumer groups to handle non idempotent operations

MahmoudSamir101:
fix issue #1148


yang-xiaodong (approved):
Looks good,  Thanks

PR #1152, opened 2022-06-10
optimizing query performance when using mysql

akinix:
in `MySqlDataStorage.cs`
``` csharp
private async Task<IEnumerable<MediumMessage>> GetMessagesOfNeedRetryAsync(string tableName)
{
    var fourMinAgo = DateTime.Now.AddMinutes(-4);
    var sql =
        $"SELECT `Id`,`Content`,`Retries`,`Added` FROM `{tableName}` WHERE `Retries`<@Retries " +
        $"AND `Version`=@Version AND `Added`<@Added AND (`StatusName` = '{StatusName.Failed}' OR `StatusName` = '{StatusName.Scheduled}') LIMIT 200;";
   ...
}
```

``` mysql
EXPLAIN
SELECT `Id`, `Content`, `Retries`, `Added`
FROM `cap.received`
WHERE `Retries` < 3
  AND `Version` = 'v1'
  AND `Added` < '2022-06-09T16:31:07.7105934+08:00'
  AND (`StatusName` = 'Failed' OR `StatusName` = 'Scheduled')
LIMIT 200
```

before optimize
``` sql
+--+-----------+------------+----------+----+-------------+----+-------+----+------+--------+-----------+
|id|select_type|table       |partitions|type|possible_keys|key |key_len|ref |rows  |filtered|Extra      |
+--+-----------+------------+----------+----+-------------+----+-------+----+------+--------+-----------+
|1 |SIMPLE     |cap.received|NULL      |ALL |NULL         |NULL|NULL   |NULL|362766|0.21    |Using where|
+--+-----------+------------+----------+----+-------------+----+-------+----+------+--------+-----------+

```
after optimized
``` sql
+--+-----------+------------+----------+-----+-----------------------------------+-----------------------------------+-------+----+------+--------+---------------------+
|id|select_type|table       |partitions|type |possible_keys                      |key                                |key_len|ref |rows  |filtered|Extra                |
+--+-----------+------------+----------+-----+-----------------------------------+-----------------------------------+-------+----+------+--------+---------------------+
|1 |SIMPLE     |cap.received|NULL      |range|IX_Retries_Version_Added_StatusName|IX_Retries_Version_Added_StatusName|5      |NULL|181383|0.63    |Using index condition|
+--+-----------+------------+----------+-----+-----------------------------------+-----------------------------------+-------+----+------+--------+---------------------+

```

### DIFF
--- a/src/DotNetCore.CAP.MySql/IStorageInitializer.MySql.cs
+++ b/src/DotNetCore.CAP.MySql/IStorageInitializer.MySql.cs
@@ -60,7 +60,8 @@ CREATE TABLE IF NOT EXISTS `{GetReceivedTableName()}` (
   `ExpiresAt` datetime DEFAULT NULL,
   `StatusName` varchar(50) NOT NULL,
   PRIMARY KEY (`Id`),
-  INDEX `IX_ExpiresAt`(`ExpiresAt`)
+  INDEX `IX_ExpiresAt`(`ExpiresAt`),
+  INDEX `IX_Retries_Version_Added_StatusName`(`Retries`, `Version`, `Added`, `StatusName`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
 
 CREATE TABLE IF NOT EXISTS `{GetPublishedTableName()}` (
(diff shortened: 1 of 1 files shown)


yang-xiaodong:
Hello, Thanks for your PR!

Because adding an index will affect write performance, so we do not add an index by default, but leave this choice to the user who needs it

PR #1154, opened 2022-06-13
Added support for GCP PubSub and Google Spanner database

jhakp:
Extended the code and added support for GCP Pub/sub (message broker) and Google Spanner database.
Added a sample application "Sample.GcpPubSub.GoogleSpanner" to validate the changes. Open appsettings.json of "Sample.GcpPubSub.GoogleSpanner" and change the PubSub details and also change the spanner connection string to try out the changes.


yang-xiaodong:
Hello,

As far as I know, Pub/Sub does not support multiple topics attached to one subscription, so we can't provide support for  CAP group,  Are you try to solve this?

jhakp:
yes, Pub/Sub does not support multiple topics to one subscription. CAP group would not be supported but we can work without that.

yang-xiaodong:
Since this is a core feature, it would be difficult for us to merge it into our main repository without support group feature, I would suggest that this could be written into our docs as a community supported extension and you as the maintainer, what do you think? 

CCommunity supported list docs: 
https://cap.dotnetcore.xyz/user-guide/en/transport/general/#community-supported-transport-extensions

Thank you.

jhakp:
I Yang, Google Pub/Sub would never support subscription group but that doesn't mean CAP cannot support GCP Pub/Sub. What Google recommends is to have only one subscription per Topic.
If you look at the code, we are registering only one subscriber. I don't think this would break any existing CAP implementation as the changes are local to the GCP Pub/Sub package.
I would sincerely like this change to be merged to master.

praveenkrjha:
Any thoughts on merging this pull request to master?

src/DotNetCore.CAP.GcpPubSub/CAP.Options.Extensions.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class CapOptionsExtensions
+    {
+        /// <summary>
+        /// Configuration to use Google Cloud Pub/Sub in CAP.
+        /// </summary>
+        /// <param name="options">CAP configuration options</param>
+        /// <param name="projectId">The GCP <c>Project</c> ID.</param>
+        public static CapOptions UseGooglePubSub(this CapOptions options,string projectId)
```
WeihanLi, line 17:
```suggestion
        public static CapOptions UseGooglePubSub(this CapOptions options, string projectId)
```

Issue #98, opened 2018-03-16
is there any method or extension allow me use typed message publish and subscribe?


Eilyyyy:
if is possible to publish and subscribe event like this ?
 ```
    public interface ICapSubscribe<in T> : ICapSubscribe
    {
        Task Handle(T message);
    }
```
```
    public class TestHandle : ICapSubscribe<TestEvent>
    {
        public Task Handle(TestEvent message)
        {
            Console.WriteLine(message.Message);
            return Task.CompletedTask;
        }
    }
````
```
    public class TestEvent
    {
        public string Message { get; set; }
    }
```
and i can publish event like this `_publisher.PublishAsync(new TestEvent())`
and the routing key just as type name, in this example is `TestEvent`.

i try to implement the interface of 'IConsumerServiceSelector' but it's internal and i found that `ConsumerExecutorDescriptor` `ConsumerContext` all is internal, seems like i have no way to replace the implement of subscribe discovery.
is there any idea to achieve this ?
thank you.

yuleyule66:
Thanks for your proposal, I will consider this later.

yuleyule66:
Hello, 

After a long time of feedback, we think that using a strongly typed subscribe interface would reduce the flexibility of application. 

Sometimes, strong typing can make us make fewer mistakes, but it's not always good. In CAP we can't ask consumers to use the same data type as the sender. Instead, they only need to be able to guarantee that these objects can be properly deserialized and they can bring a lot of flexibility. For example, you don't need to create a separate consumer class, and you can easily integrate it into the current Controller and Services.

I hope the above is useful to you, I am willing to listen to your opinion, let me konw if you have any questions.  😃

yang-xiaodong:
https://github.com/dotnetcore/CAP/tree/master/samples/Sample.RabbitMQ.SqlServer.DispatcherPerGroup/TypedConsumers

Here is an example of typed consumer

PR #1174, opened 2022-07-12
Publish and subscribe support for specified topic types

yuxingz:
Easier way to publish/subscribe.

//Eg:
```csharp
public class PublishController : Controller
{
    private readonly ICapPublisher _capBus;

    public PublishController(ICapPublisher capPublisher)
    {
        _capBus = capPublisher;
    }

    [Route("~/simple_publish")]
    public IActionResult SimplePublish()
    {
        //your business logic code

        _capBus.Publish<TopicType>(DateTime.Now);

        return Ok();
    }

    [NonAction]
    [CapSubscribe(typeof(TopicType))]
    public void CheckReceivedMessage(DateTime datetime)
    {
        Console.WriteLine(datetime);
    }
}

public class TopicType
{
}
```

### DIFF
--- a/src/DotNetCore.CAP/CAP.Attribute.cs
+++ b/src/DotNetCore.CAP/CAP.Attribute.cs
@@ -17,6 +17,10 @@ namespace DotNetCore.CAP
 
         }
 
+        public CapSubscribeAttribute(Type type, bool isPartial = false) : base(type.FullName.ToLowerInvariant(), isPartial)
+        {
+        }
+
         public override string ToString()
         {
             return Name;
--- a/src/DotNetCore.CAP/CAP.Attribute.cs
+++ b/src/DotNetCore.CAP/CAP.Attribute.cs
@@ -17,7 +17,7 @@ namespace DotNetCore.CAP
 
         }
 
-        public CapSubscribeAttribute(Type type, bool isPartial = false) : base(type.FullName.ToLowerInvariant(), isPartial)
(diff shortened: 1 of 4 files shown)


yuxingz:
This makes it easier to view pub/sub for that topic type.
It can also reduce the problems caused by inconsistent topic names.

yuxingz:
Ok,But i think publish subscribe can be simpler.
Can you think about it?
```csharp
//Eg:
public class PublishController : Controller
{
    private readonly ICapPublisher _capBus;

    public PublishController(ICapPublisher capPublisher)
    {
        _capBus = capPublisher;
    }

    [Route("~/simple_publish")]
    public IActionResult SimplePublish()
    {
        //your business logic code

        _capBus.Publish(new TopicMessageType());

        return Ok();
    }

    [NonAction]
    [CapSubscribe(typeof(TopicMessageType))]
    public void CheckReceivedMessage(TopicMessageType message)
    {
        Console.WriteLine(message.ToString());
    }
}

public class TopicMessageType
{
}
```

yang-xiaodong:
You only consider the object type, also need to consider other types, such as

_capbus.publish (123);
_capbus.publish ("bar");
_capbus.publish (new byte []);

yuxingz:
> No, simple types don't need to be used this way, because the topic names are based on the complex type's namespace.

Thanks your reply, I referenced #98 , plan to implement it in the
my solution.

PR #1191, opened 2022-08-25
Added BeginTransaction override with possibility to specify Transaction isolation level for SQL Server

ewgenym:
A helpful override when SQL Server storage is used and business transaction requires a specific transaction isolation level.


yang-xiaodong:
Hello,

Thanks for your PR. 

I'm sorry to tell you that we are not going to add this override method, users can add extension methods themselves to do this, and the PR only for SQLServer, different storages need the same behavior, so I will close the PR! Thanks

PR #1193, opened 2022-08-27
Fixes MongoDB queries not returning results when a element convention name is registered

rpenha:
When CAP is using MongoDB with a different element name convention registered in the convention registry, the queries does not work as expected.

This occurs because some queries are mapping properties as strings, and once a different case convention is registered (CamelCaseElementNameConvention, as example), they don't behave as expected.

In order to resolve this issue, I changed the queries to use LINQ instead of Bson Documents. I didn't try to abstract PublishedMessage and ReceivedMessage and simply encapsulate the necessary logic to behave accordingly in each case based on the message type.

### DIFF
--- a/samples/Sample.RabbitMQ.MongoDB/docker-compose.yml
+++ b/samples/Sample.RabbitMQ.MongoDB/docker-compose.yml
@@ -1,0 +1,41 @@
+version: "3.9"
+services:
+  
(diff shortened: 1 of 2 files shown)


yang-xiaodong (approved):
Looks Good, Thanks

Issue #1024, opened 2021-10-28
Question about batch sending


naruto1227:
I would like to ask why cap after restart after reading messages from the database is serial execution

![image](https://user-images.githubusercontent.com/43592384/139181493-413d862f-9eb0-455c-8890-8c14753fb13d.png)

PR #1201, opened 2022-09-06
fix Consumer Service register error when with scope lifetime

sampsonye:
if service or it's dependent services register with scop ， it can not resolve from root ServiceProvider
image


yang-xiaodong:
Hi, please check the Issue's  #1024 comment

sampsonye:
> Hi, please check the Issue's #1024 comment

it related to #1204   I add some more info

yang-xiaodong (approved):
LGTM,  Thanks

PR #1205, opened 2022-09-11
Support .NET Framework and .NET Standard 2.0

bridget-spotlightreporting:
Make the CAP library usable from .NET Framework projects. Also, a minor change so CAP doesn't get infinitely stuck if the incoming queue contains messages without CAP headers.

Changes:
- Update csproj files' target frameworks, and also use the target-specific version of Entity Framework (note, database projects explicitly support net472 because there seems to be an issue in nuget that means framework projects think they need the .NETStandard2.1 dependencies unless there's a framework-specific version of the nuget package)
- Remove\pragma C#8+ features
    - [NotNullWhen] tags
    - Methods implementations defined on the interface
    - Pragma around the different features available in different EntityFramework versions
    - Add pragmas to suppress unavoidable warnings (e.g. that the NetStandard2.0 version of a method has no async content)
- Add alternative way to get IBootstrapper running in the background (since .NET Framework just ignores IHostedService)
    - Make Bootstrapper class public and add StartIfNotAlreadyRunning method and support
    - Let ConsumingAssembly be specified in options because Assembly.EntryAssembly() doesn't find the right result when Bootstrapper triggered manually
- Add sample .NET Framework project
- Add null checks around getting headers on a received message

### DIFF
--- a/CAP.sln
+++ b/CAP.sln
@@ -86,7 +86,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Sample.RabbitMQ.SqlServer.D
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DotNetCore.CAP.OpenTelemetry", "src\DotNetCore.CAP.OpenTelemetry\DotNetCore.CAP.OpenTelemetry.csproj", "{83DDB126-A00B-4064-86E7-568322CA67EC}"
 EndProject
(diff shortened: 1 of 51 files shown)


bridget-spotlightreporting:
Hi,

We're using this library for a project (thanks!) but needed to make it support .NET Framework for one of the applications, so wondered if you'd be interested in this too.

Thanks,
Bridget

bridget-spotlightreporting:
Hi,

Fair enough, .NET Framework is getting a bit old now. Feel free to reference us as community support.

Thanks,
Bridget

Farawa-y:
@bridget-spotlightreporting Hello friend, excuse me.  sqlserver+rabbitmq, is framewrok4.8 supported ?

bridget-spotlightreporting:
Hi @Farawa-y - it's not something we've tested, but theoretically should work - our fork is at https://github.com/spotlight-reporting/CAP, feel free to try it out and see. 

Please note, this is not an actively maintained fork - we forked from this main dotnetcore project at version 2.1.1 and have not kept up to date since, so none of the features developed by Yang since then are on our fork. If the fork works for you as-is, then that's great.

PR #1212, opened 2022-09-22
When using RedisStreams only one ConsumerGroup is created 

mlatoszek:
Fixed RedisStream.Manager.Extensions.TryGetOrCreateStreamGroupAsync to create ConsumerGroup when not found.

### DIFF
--- a/build/version.props
+++ b/build/version.props
@@ -2,7 +2,7 @@
 	<PropertyGroup>
 		<VersionMajor>6</VersionMajor>
 		<VersionMinor>2</VersionMinor>
-		<VersionPatch>0</VersionPatch>
+		<VersionPatch>1</VersionPatch>
 		<VersionQuality></VersionQuality>
 		<VersionPrefix>$(VersionMajor).$(VersionMinor).$(VersionPatch)</VersionPrefix>
 	</PropertyGroup>
--- a/src/DotNetCore.CAP.RedisStreams/IRedisStream.Manager.Extensions.cs
+++ b/src/DotNetCore.CAP.RedisStreams/IRedisStream.Manager.Extensions.cs
@@ -66,6 +66,9 @@ namespace DotNetCore.CAP.RedisStreams
                 var groupInfo = await database.StreamGroupInfoAsync(stream);
                 if (groupInfo.Any(g => g.Name == consumerGroup))
                     return;
+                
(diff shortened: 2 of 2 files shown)


mlatoszek:
When using RedisStreams as a transport along with multiple Consumer Groups - the current implementation throws exception when staring subscriber that has different consumer group that the one that created the stream. It seems like creation part was missing from ```TryGetOrCreateStreamGroupAsync``` method.

MahmoudSamir101:
@mlatoszek , Could you please confirm if I understood correctly the issue, you had two deployed instances that use CAP RedisStreams , and each one of them has a unique DefaultGroupName, or do both instances use the same name? 
as if they use different names, this is already handled when reading the stream group by StreamName and then filtering the group by DefaultGroupName
after that we are creating the group if it was not there, you will find that in _TryGetOrCreateStreamConsumerGroupAsync_ method the public one

mlatoszek:
Hi @MahmoudSamir101,

The issue occurs when there are two instances with different consumer group name - using the same stream. Please see the code below:
image
if (streamExist) the second consumer group will not be created as the operations ends on checking if there is already a consumer group will the same name registered in the stream - but there is - nothing happens. 

That's why TryGetOrCreateStreamGroupAsync IMO should create stream group if it does not exist - even name of the method suggest it.

MahmoudSamir101:
@mlatoszek , you are right, we do nothing if a stream was created from one of the consumer groups
Thanks for the fix

mlatoszek:
My pleasure @MahmoudSamir101  :) 
Btw. do you know when can it be merged and the fix released with new nugets? I need it in order to use CAP in our project. I can create nugets  myself and release it to our private repo but would be great to use official ones ;)

yang-xiaodong:
Hi @MahmoudSamir101 ,
You can review and merge this PR directly, and I will push the 6.2.1 preview verion to nuget

MahmoudSamir101:
@yang-xiaodong should I keep the **_version.props_** changes or you will manage the version number

yang-xiaodong:
> should I keep the version.props changes

Yes, that's exactly our next version number

MahmoudSamir101 (commented):
Creating consumer group if it is not created

yang-xiaodong:
Fixed in version `6.2.1-preview-180716003`  and pushed to nuget in a few minutes ago.

PR #1220, opened 2022-10-13
Update SnowflakeId.cs

cuibty:
EnvironmentVariableTarget.Machine 

The environment variable is stored or retrieved from the HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment key in the Windows operating system registry. This value should be used on .NET implementations running on Windows systems only.

https://learn.microsoft.com/en-us/dotnet/api/system.environmentvariabletarget?view=net-6.0


yang-xiaodong (approved):
LGTM, Thanks

PR #1227, opened 2022-10-26
The case of parameter status and database statusName is inconsistent,…

DaviaWang:
… resulting in no query results.(Cases from Mysql)


yang-xiaodong:
This is intentional, the  state statusName is case sensitive here

image

Issue #1232, opened 2022-11-03
RabbitMQ cluster connection failed without using default ports


yibei333:
**环境：**
```
<Project Sdk="Microsoft.NET.Sdk.Web">
	<PropertyGroup>
		<TargetFramework>net6.0</TargetFramework>
		<ImplicitUsings>enable</ImplicitUsings>
	</PropertyGroup>
	<ItemGroup>
		<PackageReference Include="DotNetCore.CAP.SqlServer" Version="6.2.1" />
		<PackageReference Include="DotNetCore.CAP.RabbitMQ" Version="6.2.1" />
		<PackageReference Include="DotNetCore.CAP.Dashboard" Version="6.2.1" />
		<PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="6.0.10" />
		<PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="6.0.10">
		  <PrivateAssets>all</PrivateAssets>
		  <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
		</PackageReference>
	</ItemGroup>
</Project>
```

**配置代码:**
```
builder.Services.AddCap(options =>
{
    options.UseEntityFramework<FooDbContext>();
    options.UseRabbitMQ(x =>
    {
        //如果写x.HostName="localhost"没有问题，多个逗号分割就会报错
        x.HostName = "localhost:5672,localhost:5673,localhost:5674";
        x.UserName = "user";
        x.Password = "pass";
    });
    options.UseDashboard();
});
```

**错误日志：**
```
fail: DotNetCore.CAP.RabbitMQ.ConnectionChannelPool[0]
      RabbitMQ channel model create failed!
      RabbitMQ.Client.Exceptions.BrokerUnreachableException: None of the specified endpoints were reachable
       ---> System.AggregateException: One or more errors occurred. (Connection failed) (Connection failed) (Connection failed)
       ---> RabbitMQ.Client.Exceptions.ConnectFailureException: Connection failed
       ---> System.Net.Sockets.SocketException (11001): No such host is known.
         at System.Net.NameResolutionPal.ProcessResult(SocketError errorCode, GetAddrInfoExContext* context)
         at System.Net.NameResolutionPal.GetAddressInfoExCallback(Int32 error, Int32 bytes, NativeOverlapped* overlapped)
      --- End of stack trace from previous location ---
[...]

PR #1233, opened 2022-11-05
support connect to RabbitMQ cluster without using default port

yibei333:
reference the issue #1232


yang-xiaodong (approved):
LGTM, Thanks

PR #1239, opened 2022-11-22
update general.md for add MQTT extensions support

jinzaz:
Add the connection guide of the extended warehouse


yang-xiaodong (approved):
Looks good, Thanks

PR #1253, opened 2022-12-28
Change AzureServiceBus nuget package from Microsoft.Azure.ServiceBus to Azure.ServiceBus

giorgilekveishvili-meama:
Removed Microsoft.Azure.ServiceBus implementation and added Azure.ServiceBus


yang-xiaodong:
Thanks for the PR!

Hi @mviegas , Can you review this PR? I don't know much about the details of ASB

mviegas:
Hi @yang-xiaodong <https://github.com/yang-xiaodong>!

I'm currently on vacation with mobile-only access to internet. Let me get
back to you on this PR in January 2nd?!

Thanks!

yang-xiaodong:
@mviegas  OK, Great!
Thanks!

src/DotNetCore.CAP.AzureServiceBus/AzureServiceBusConsumerClient.cs
```diff
@@ -78,45 +81,51 @@ public void Listening(TimeSpan timeout, CancellationToken cancellationToken)
...
-            var commitInput = (AzureServiceBusConsumerCommitInput) sender!;
-            if (_asbOptions.EnableSessions)
-            {
-                commitInput.Session?.CompleteAsync(commitInput.LockToken);
-            }
-            else
-            {
-                _consumerClient!.CompleteAsync(commitInput.LockToken);
-            }
+            //var commitInput = (AzureServiceBusConsumerCommitInput) sender!;
```
mviegas, line 120: Hey @giorgilekveishvili-meama. Could you clarify why is this code here commented? If the `AutoCompleteMessages` property is set to false, the message must be manually completed. Check [this section on the Migration Guide from Microsoft](https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-transfers-locks-settlement#settling-receive-operations).

Thank you!

src/DotNetCore.CAP.AzureServiceBus/CAP.AzureServiceBusOptions.cs
```diff
@@ -25,6 +24,11 @@ public class AzureServiceBusOptions
         /// </summary>
         public string ConnectionString { get; set; } = default!;
 
+        /// <summary>
+        /// Namespace of Servicebus
+        /// </summary>
+        public string Namespace { get; set; } = default!;
```
mviegas, line 30: Hi! For these new properties you introduced here, it would be good to add some comment explaining their purpose and also [update the documentation with it](https://github.com/dotnetcore/CAP/blob/master/docs/content/user-guide/en/transport/azure-service-bus.md).

src/DotNetCore.CAP.AzureServiceBus/CAP.AzureServiceBusOptions.cs
```diff
@@ -36,14 +40,18 @@ public class AzureServiceBusOptions
         /// </summary>
         public string TopicPath { get; set; } = DefaultTopicPath;
 
+        public bool AutoCompleteMessages { get; set; }
```
mviegas, line 43: Same as comment above.

src/DotNetCore.CAP.AzureServiceBus/CAP.AzureServiceBusOptions.cs
```diff
@@ -36,14 +40,18 @@ public class AzureServiceBusOptions
...
+
+        public int MaxConcurrentCalls { get; set; }
```
mviegas, line 45: Same as comment above.

mviegas (changes requested):
Looks great, thanks for this! I just put a few comments regarding:

- Documentation of new properties in AzureServiceBusOptions;
- Completion of messages when `AutoCompleteMessages` is set to false;

giorgilekveishvili-meama:
@mviegas  Please re-check

src/DotNetCore.CAP.AzureServiceBus/AzureServiceBusConsumerClient.cs
mviegas: Just adding for registration: the `Commit` method is called [here](https://github.com/dotnetcore/CAP/blob/master/src/DotNetCore.CAP/Internal/IConsumerRegister.Default.cs) when storing a message in the Inbox.

mviegas:
> @mviegas Please re-check

Thanks for resolving the Commit method. Once the other comments regarding properties documentation (in class and in public docs) are solved, we can merge this.

giorgilekveishvili-meama:
I have added comments and updated documentation

mviegas (approved):
Looks good to go for me, @yang-xiaodong . Thanks for solving the questions @giorgilekveishvili-meama !

yang-xiaodong:
Good, thanks guys

PR #1266, opened 2023-01-27
Add DatabaseFacade BeginTransaction overload with IsolationLevel argument and async version

xshaheen:


yang-xiaodong (approved):
LGTM, Thanks

yang-xiaodong:
Note:
Since Transaction is implemented based on AsyncLocal, it does not support passing in an asynchronous context to parent, so we will remove BeginTransactionAsync in the next version

PR #1267, opened 2023-02-03
add basicQos options to control the basic channel of RabbitMq and enforce qos parameters

nunorelvao:
feature: add basicQos options to control the basic channel of RabbitMq and enforce qos parameters.

This feature is a much needed option and a simple one to implement for RabbitMQ control of the basicQos channel.

The reason why the need for this setup:in my opinion is a much needed option due to the fact that sometimes we need to control the flow of the messages consumed to allow for lesser memory footprint on some big queues with some big messages workload that cannot be coped with the rabbit speed of delivery to consume those messages.

I would like to know if is possible to add this change also from at least 6.2.1 as there are still some projects in the .net 6 version?

Thanks to: @rafaelpadovezi / @ricardochaves  helping sorting this out

### DIFF
--- a/src/DotNetCore.CAP.RabbitMQ/CAP.RabbiMQOptions.cs
+++ b/src/DotNetCore.CAP.RabbitMQ/CAP.RabbiMQOptions.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading.Channels;
 using RabbitMQ.Client;
 using RabbitMQ.Client.Events;
 
@@ -79,7 +80,7 @@ namespace DotNetCore.CAP
         /// Optional queue arguments, also known as "x-arguments" because of their field name in the AMQP 0-9-1 protocol,
         /// is a map (dictionary) of arbitrary key/value pairs that can be provided by clients when a queue is declared.
         /// </summary>
-        public QueueArgumentsOptions QueueArguments { get; set; } = new ();
+        public QueueArgumentsOptions QueueArguments { get; set; } = new();
 
         /// <summary>
         /// If you need to get additional native delivery args, you can use this function to write into <see cref="CapHeader"/>.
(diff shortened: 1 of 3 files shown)


yang-xiaodong:
Hi @nunorelvao 

It's make sense, I will review the code later , thanks for your PR!

yang-xiaodong:
Before version 7.0, we'll prefeth some message into memory queue and then distribute to dispatcher to execution. Which will also cause some issues while improving performance.

For example, when the subscription method is executed too slowly for too long, it will lead the retry thread to get the schedule status message that has not yet been executed. ( the retry thread default pick message 4 minutes ago)

So in version 7.0 we have changed the default behavior. Now, by default, only one message will be read from the message queue, and then the subscription method will be executed. After the execution is completed, the next message will be read for execution. We added `EnableConsumerPrefetch` option to use the previous behavior. The default is False now.

nunorelvao:
Oh , thanks i understand now why i was seeing some duplication on some very
long queue and high memory consumption on the consumer side. So this
combined with version 7 and this new options i am  proposing we should be
able to have much better control if needed because we can control the
basicqos. Please review it and adjust if you think can be better handled.

src/DotNetCore.CAP.RabbitMQ/CAP.RabbiMQOptions.cs
```diff
@@ -104,5 +116,50 @@ public class QueueArgumentsOptions
...
+                PrefetchSize = prefetchSize;
+                PrefetchCount = prefetchCount;
+                Global = global;
+            }            
+
+            /// <summary>
+            /// Gets the PrefetchSize, a value of 0 is treated as infinite, allowing any number of unacknowledged message to be prefetched on server. 
+            /// The default value is 0.
+            /// </summary>
+            public uint PrefetchSize { get; private set; } = 0;
```
yang-xiaodong, line 149: Hi @nunorelvao ,

I noticed that the prefetch_size option is not implemented in RabbitMQ, I think we can remove it.

https://groups.google.com/g/rabbitmq-users/c/iiv_HPB-nqI

yang-xiaodong (changes requested):

src/DotNetCore.CAP.RabbitMQ/CAP.RabbiMQOptions.cs
yang-xiaodong: https://www.rabbitmq.com/specification.html

image

nunorelvao: Yes you right and agreed , making this change to remove this option.

yang-xiaodong (approved):
LGTM,  Thanks for the PR!

Issue #1273, opened 2023-02-18
a better way to acquire storage lock


li-zheng-hao:
The current approach is to compete a single lock for all instances of services, when a instance A1 of service A get the publish or received retry lock, all instances of other services cannot retry their failure messages (actually they can and should concurrently process) , it may make some services cannot retry failure messages for a long time.

maybe we can acquire the lock according to `publishedcollection` and `receivedcollection` in storage options, in this way, the lock range is smaller, and only instances that share the same table will compete for the same lock

yang-xiaodong:
Other services and service A may use different databases or tables, why can't other services retry their own failures  message?

li-zheng-hao:
The number of services in my project is relatively small, so I only use one database. Maybe my application scenarios are very few, so just ignore the issue ……

yang-xiaodong:
How are you using multiple services in one pub/rec table?

li-zheng-hao:
no, different services use different published/ received  tables, but when we use only one database for all services, there's may be a problem, let me explain it in details

when one instance of service A acquired lock, it  will go to retry failure messages , if all intsance of service B try to acquire lock at the same time, they cannot acquire the lock, because this lock publish_retry or received_retry is hold by the instance of service A, so they will return immediately and sleep 60s

If the preceding steps are repeated for many times, the failure message of service B cannot be retried for a long time

of course, none of this is a problem if different services use different databases

is there any wrong with my explanation?

yang-xiaodong:
Are you use `TableNamePrefix` option to put all service cap table in same database?
```
serviceA.published
serviceA.received
serviceA.lock
```
```
serviceB.published
serviceB.received
serviceB.lock
```

yang-xiaodong:
I realized this evening that we overlooked a potentially problematic scenario.

When using the `Version` configuration option to isolate data, it may cause problems because the lock table does not have a Version field for isolation.   We don't have to add a Version column in the lock table, just change the Key filed.

Lock table key filed:
publish_retry --> publish_retry+capOption.Version
received_retry --> received_retry +capOption.Version

Add 2 internal fields in CapOptions.cs, and use these 2 fields in the SQL operation Key, Would you like to submit a PR to fix it? 

![image](https://user-images.githubusercontent.com/8205994/219876303-28a603e3-0d2d-4012-9f1e-b6406e10a49f.png)

li-zheng-hao:
sure！I'm more than happy to help make CAP better

[...]

PR #1274, opened 2023-02-19
fix: append version option to lock key

li-zheng-hao:
solve a potentially problematic scenario in this issue: https://github.com/dotnetcore/CAP/issues/1273


li-zheng-hao:
maybe we dont need to add two internal option `LockPubKey` and `LockRecKey`,  because they are only used in `IProcessor.NeedRetry.cs` ,

we can just use `$"received_retry_{_capOptions.Value.Version}"` and `$"publish_retry_{_capOptions.Value.Version}"` , This makes the `CapOption.cs` class look cleaner

what do you  think about this? @yang-xiaodong

yang-xiaodong:
Yes, i agree with you, just do it

yang-xiaodong (approved):
LGTM, Thanks for your PR!

PR #1281, opened 2023-02-24
Added option to set the queue type for RabbitMQ

PaulCousinsTTEducation:
Added an option to change the consumer queue type if required as I needed support for quorum queues.

https://www.rabbitmq.com/quorum-queues.html


yang-xiaodong (approved):
LGTM, Thanks for your PR!

Issue #1292, opened 2023-03-14
file scoped namespaces


Revazashvili:
currently project is partially using file scoped namespaces. 
only several projects: DotNetCore.CAP, DotNetCore.CAP.SqlServer and DotNetCore.CAP.AzureServiceBus.Test are converted to file scoped namespaces, 
should not all projects be converted to file scoped namespaces?

yang-xiaodong:
Hello, this is not urgent and important, maybe a change will be made in a major version in the future

PR #1294, opened 2023-03-17
convert solution to file scoped namespaces

Revazashvili:


Revazashvili:
it will be changed in future versions #1292

PR #1295, opened 2023-03-17
azure 可配置客户端传输类型

dot-lemon:
这个功能比较紧急，能否在明天发布新版本


yang-xiaodong:
Please use English and describe the motivation of this feature

dot-lemon:
Sorry, I haven't modified the function here correctly. I'll close this PR first. Currently, Azure Service Bus communication is used using WebSocket, and only the default TCP in the configuration makes it impossible to publish subscriptions. Then, when subscribing, an administrator client will be created, and my lack of permissions will cause errors when creating a topic.

Issue #1304, opened 2023-03-28
Dashboard and JWT Authentication


dima-zhemkov:
Hello,

I use JWT authentication and I don't use Cookie nor OpenID connect and it makes me almost impossible to use CAP Dashboard with authentication.

It would be awesome if dashboard home page accepted query parameter and saved them to local storage.

For instance: 
After opening URL `https://mydomain.com/cap/index.html?token=tokenValue...&refreshToken=refreshtokenValue...`
two local storage entries will be added/updated:
```
token = tokenValue...
refreshToken = refreshtokenValue...
```
After that send them in headers with each request.

mviegas:
Hi @dima-zhemkov !

You can leverage ASP.NET Core flexible authentication mechanism to use any authentication and challenge schemes, including custom ones. Please look at the [Custom authentication section](https://cap.dotnetcore.xyz/user-guide/en/monitoring/dashboard/#custom-authentication) in the docs.

dima-zhemkov:
Hi @mviegas!

I've tried it.

It doesn't work without cookies since CAP Dashboard doesn't send auth headers with requests

yang-xiaodong:
Sounds reasonable, since our current `/cap/index.html` needs to be authenticated first, so we need to detect when the access_token is passed from query we need to cancel the authentication for index.html and write token to localstorage.

dima-zhemkov:
Will work on it and prepare PR shortly.

yang-xiaodong:
Fixed in `7.1.1-preview-197551401`

PR #1306, opened 2023-03-31
Log in to dashboard with JWT authentication

dima-zhemkov:
Implemented ability to receive access token as URI query parameter in compliance with  [RFC6750](https://www.rfc-editor.org/rfc/rfc6750#section-2.3).

Solution for #1304


yang-xiaodong (approved):
LGTM ,Thanks

Issue #1307, opened 2023-04-04
dashboard nodes switch failure


wwwu:
It is always the default node data

yang-xiaodong:
Fixed in `7.1.1-preview-197551401`

PR #1308, opened 2023-04-04
Fix dashboard nodes switch failure. #1307

wwwu:


yang-xiaodong:
Thanks the PR. Through my test the PR couldn't work. I have created a new commit to fix the issue. Thanks for your feedback.

Issue #1322, opened 2023-04-29
Change the SnowflakeId from static singleton to dependency injection singleton


3ldar:
Hi,
We are getting errors like below while both publishing and receiving events.

>SubscriberExecutionFailedException-->23505: duplicate key value violates unique constraint "published_pkey" DETAIL: Key ("Id")=(1613554411973534) already exists."

We run our application in a k8n cluster with 10 or more pods distributed to multiple nodes. (We use openshift 3.11 version if it helps). To my guess it is due to the `SnowflakeId` implementation. It uses mac addresses last 10 bits to generate a `workerId` if it is not provided in environmental variable `CAP_WORKERID`. There is a possibility that those 10 bits may be the same in a multi node cluster. In a single node cluster the OUI part is unique as the initial part is the same but when it comes to multiple nodes there is a possibility that the initial part of the mac is different but the last part is the same. When this happens and under a good load we get this kind of errors both while publising or receiving.  It is an option to set `CAP_WORKERID` for each pod but openshift gives you the only one environmental configuration per deployment config not per pod. (It doesn't make sense having multiple deployment configs for the same service as the number is dynamic due to load etc.) 

I have created a ugly  workaround like below. I have used redis to increment a number to assign as workerId and set it to  `CAP_WORKERID` env value.

```
private const string CapWorkerIdKey = "CAP_WORKERID";
private static void ConfigureWorkerId(IServiceCollection services)
        {
            var sp = services.BuildServiceProvider();
            var redis = sp.GetRequiredService<IRedisCacheClient>();
            var db = redis.Db3.Database;

            var workerId = db.StringIncrement(CapWorkerIdKey, 10);
            db.KeyExpire(CapWorkerIdKey, TimeSpan.FromMinutes(5));
            Environment.SetEnvironmentVariable(CapWorkerIdKey, workerId.ToString()); // Need to set EnvironmentTarget.Machine in older  versions like 6.x
        }
```

It seems to work for me. But a singleton service would be nicer to have for `SnowflakeId` where I can fiddle the `workerId` generation or entire `SnowflakeId` generation process.

yang-xiaodong:
This has nothing to do with a single-node cluster or a multi-node cluster, the mac address is created and assigned by Docker.

[...]

PR #1328, opened 2023-05-05
Change the SnowflakeId from static singleton to dependency injection singleton #1322

3ldar:
### Description:
Converts `SnowflakeId` from a singleton object into a singleton service that can be altered by the users.

#### Issue(s) addressed:
- #1322

#### Changes:
- It might introduce a breaking change if someone is using the `SnowflakeId.Default()` method.

#### How to test:
- In the startup, after calling `AddCap` to register cap services a user can add a singleton service that implements `ISnowlakeId` implements to change how the cap message Ids are generated.

### Checklist:
- [ ] I have added the necessary documentation (if applicable)

### Reviewers:
- @yang-xiaodong


yang-xiaodong:
The CustomHeaders option is present in almost every Transport, and I see that you added a `CustomHeadersBuilder` in AzureServiceBus, It doesn't look pretty. so I'm wondering if it's worth making such a big change for this? and you now using Redis to set environment variables doesn't looks bad, let me know what do you think?

3ldar:
> Hi @3ldar ,
> 
> Thanks for your PR!
> 
> The CustomHeaders option is present in almost every Transport, and I see that you added a `CustomHeadersBuilder` in AzureServiceBus, It doesn't look pretty. so I'm wondering if it's worth making such a big change for this? and you now using Redis to set environment variables doesn't looks bad, let me know what do you think?

I know it works and serves the purpose. But as a general design note, I don't like to have singleton-like anti-patterns in my code where I can have a singleton service instead. To me, it's worth making all the `CustomHeadersBuilder`s to provide an `IServiceProvider` instance because in some scenarios you will eventually need some services (Like a DbContext, or a business service) and to be able to make it work you will need to build services to get the service provider, which is not recommended and also will mess up your already registered singleton services, besides it might be even impossible when a service depends on a cap service (like `ICapPublisher`) and you need that service for the custom headers generation. So will add an alternate for all the `CustomHeaders` in the options.

src/DotNetCore.CAP.AzureServiceBus/CAP.AzureServiceBusOptions.cs
```diff
@@ -62,13 +65,22 @@ public class AzureServiceBusOptions
...
         public Func<ServiceBusReceivedMessage, List<KeyValuePair<string, string>>>? CustomHeaders { get; set; }
 
+        /// <summary>
+        /// Use this function to write additional headers from the original ASB Message or any Custom Header, i.e. to allow compatibility with heterogeneous systems, into <see cref="CapHeader"/>       
+        /// </summary>
+        /// <remarks>
+        /// This property is a copy cat of <see cref="CustomHeaders"/>. The difference is: It also accpets a <see cref="IServiceProvider"/> to use DI services.
+        /// This is added due to converting <see cref="SnowflakeId"/> into a singleton service.
+        /// </remarks>
+        public Func<ServiceBusReceivedMessage, IServiceProvider, List<KeyValuePair<string, string>>>? CustomHeadersBuilder { get; set; }
```
yang-xiaodong, line 75: I suggest Add `[Obsolete("xxxxx")]` for old `CustomHeaders` option, we will  remove old option at the feature

src/DotNetCore.CAP.Kafka/CAP.KafkaOptions.cs
```diff
@@ -48,6 +48,11 @@ public KafkaOptions()
         /// </summary>
         public Func<ConsumeResult<string, byte[]>, List<KeyValuePair<string, string>>>? CustomHeaders { get; set; }
 
+        /// <summary>
+        /// If you need to get offset and partition and so on.., you can use this function to write additional header into <see cref="CapHeader"/>
+        /// </summary>
+        public Func<ConsumeResult<string, byte[]>,IServiceProvider, List<KeyValuePair<string, string>>>? CustomHeadersBuilder { get; set; }
```
yang-xiaodong, line 54: I suggest Add `[Obsolete("xxxxx")]` for old `CustomHeaders` option, we will  remove old option at the feature

src/DotNetCore.CAP.RabbitMQ/CAP.RabbiMQOptions.cs
```diff
@@ -84,6 +85,11 @@ public class RabbitMQOptions
         /// </summary>
         public Func<BasicDeliverEventArgs, List<KeyValuePair<string, string>>>? CustomHeaders { get; set; }
 
+        /// <summary>
+        /// If you need to get additional native delivery args, you can use this function to write into <see cref="CapHeader"/>.
+        /// </summary>
+        public Func<BasicDeliverEventArgs, IServiceProvider, List<KeyValuePair<string, string>>>? CustomHeadersBuilder { get; set; }
```
yang-xiaodong, line 91: I suggest Add `[Obsolete("xxxxx")]` for old `CustomHeaders` option, we will  remove old option at the feature

src/DotNetCore.CAP/Internal/SnowflakeId.Default.cs
```diff
@@ -71,30 +67,15 @@ public SnowflakeId(long workerId)
...
-    public static SnowflakeId Default()
-    {
-        if (_snowflakeId != null) return _snowflakeId;
-
-        lock (SLock)
-        {
-            if (_snowflakeId != null) return _snowflakeId;
-
-            if (!long.TryParse(Environment.GetEnvironmentVariable("CAP_WORKERID"), out var workerId))
-                workerId = Util.GenerateWorkerId(MaxWorkerId);
```
yang-xiaodong, line 85: Cannot be removed, placed into the constructor as the default behavior

src/DotNetCore.CAP/CAP.ServiceCollectionExtensions.cs
```diff
@@ -29,7 +31,7 @@ public static CapBuilder AddCap(this IServiceCollection services, Action<CapOpti
 
         services.AddSingleton(_ => services);
         services.TryAddSingleton(new CapMarkerService("CAP"));
-
+        services.AddSingleton<ISnowflakeId>(_ => new SnowflakeId(Util.GenerateWorkerId(1023)));
```
yang-xiaodong, line 34: I suggest here
```
services.AddSingleton<ISnowflakeId,SnowflakeId>();
```

src/DotNetCore.CAP/Internal/SnowflakeId.Default.cs
3ldar: Added a parameterless constructor that utilizes the above method.

yang-xiaodong (approved):
Thanks for the PR, since this is a Breaking Changes we will merge in the next minor release

PR #1369, opened 2023-07-11
Feature/asb rework

Wiks00:
### Description:
Redesign of the project:
Removed unnecessary properties from the settings object
Refactoring code to make it work as intended
Minor optimization

#### Issue(s) addressed:
- [Link to the related issue(s), if any]

#### Changes:
- Default topic and other properties that are not implemented correctly have been removed from AsbOptions
- Factory does not re-create topics and duplicate connections
- Message processors are now correctly created by the CapSubscribe attribute
- Publishing to multiple Azure Service Bus topics 
- Сorrect disposal of objects

#### Affected components:
- DotNetCore.CAP.AzureServiceBus

#### How to test:
Nothing new, the project now works like other packages related to message transport

### Checklist:

### Reviewers:


mviegas:
Hi @Wiks00! 

Thanks for submitting your PR. Could you please provide more context and motivation on the changes you're introducing here? Those need to be properly evaluated on a case by case basis, as some of them are breaking changes to the current API exposed via the Options file.

Wiks00:
Hello @mviegas. 
The previous version didn't work at all, as the most obvious one is that ServiceBusProcessor was created by default topic instead of CapSubscribe attribute for some reason. All other changes only fix the library's functionality for asb, as indicated in the description

mviegas:
Hi @Wiks00! Thanks for the feedback. 

First of all, thank you for your contribution. Unfortunately, after analysing the PR, we won't be able to move with the merge as it introduces several breaking changes to the current Options API for both default/custom producers and subscribers. I'll try to describe some of those below as well as why they cannot be merged at this time.

**About the removal of the `DefaultTopicPath` property:**

We cannot remove it as it would configure a breaking change on the current Options API. It introduces a breaking change on how the Topics are defined in both default/custom producers and subscribers. 

There are multiple away of registering subscribers in CAP, being the `CapSubscribeAttribute` one of them. You can check [this example](https://github.com/dotnetcore/CAP/blob/master/test/DotNetCore.CAP.Test/CustomConsumerSubscribeTest.cs) to see other ways like via extending the `ConsumerServiceSelector` base class or even with custom attributes. 

That said, by removing the `DefaultTopicPath` we cannot be sure on how consumer on the API are leveraging this property to build their own custom subscribers. Since there's no bug introduced by this property the way it is nowadays, we shouldn't remove it as of now as there's no enough motivation that justifies it.

**About the removal of `EnableSessions` property**:

We cannot remove this property as well, as it would introduce a breaking change on the ASB sessions feature, which is currently supported by CAP.

**About the removal of `AutoCompleteMessages` property**:

We cannot remove this property as well, it was a feature introduced in #1369.

**About the removal of `SQLFilters` property:

We cannot remove this property as well, it would introduce a breaking change on the support of this native ASB feature.

**About the change in the `IServiceBusProducerDescriptor` interface**:

Currently it is possible to define custom producers for a given type, that's why the `ConfigureCustomProducer` method in the Options API is generic method. By introducing your proposed change, we would lose this possibility and also cause breaking changes to the custom producers functionality added to this transport in #1283 .

---

Nevertheless, if there are any improvements you envision here such as:
- Connectivity enhancements;
- Introduction of new properties to the Options API;
- Bug fixing of a given functionality;

I'd recommend to address each one of them in a separate PR due to separation of concerns, so that we can analyse and address them properly.

Hope you understand, and we're open to new PRs with the constraints above now clear. 😀

Wiks00:
Hello @mviegas.
Thanks for the detailed response. As for how to move forward, could I create a few bugs describing that the above supported features does not work as specified and link them to a new merge request ? Sorry, but it just does not seem right to me to provide functionality that does not cover the functional requirements ...

mviegas:
Hi @Wiks00! 

As for "functionality that does not cover the functional requirements", we're an open source project and each contributor is the one responsible for checking the functionality since, as of now, we do not have integration tests against a real ASB instance. And I'm pretty sure that none of our contributors would want to merge a feature that just do not work. 🙂

That said, feel free to create the bugs as well to address corrections to each one of bugs you mentioned. If you could also send some "how to reproduce" it would be great, so that we can tag some contributors who originally shipped the feature to check on that as well. Thanks for your contribution!

Issue #1387, opened 2023-08-04
[System.Diagnostics] Opentelemetry and Dynatrace One Agent


phmonte:
Is there any reason to use [_contexts](https://github.com/dotnetcore/CAP/blob/0d8e9357d26a57140528141b30feb926e6a53bdb/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs#L24) instead of using [Propagator.Inject](https://github.com/dotnetcore/CAP/blob/0d8e9357d26a57140528141b30feb926e6a53bdb/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs#L111) for this function?

In the [CapEvents.BeforePublish](https://github.com/dotnetcore/CAP/blob/0d8e9357d26a57140528141b30feb926e6a53bdb/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs#L94) event, as the Context of a _contexts variable is retrieved, the [IsRemote ](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitycontext.isremote?view=net-8.0#system-diagnostics-activitycontext-isremote)property of Context is false by default, however this variable should be true([according to microsoft documentation](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitycontext.isremote?view=net-8.0#system-diagnostics-activitycontext-isremote)), as the [CapEvents.BeforePublish ](https://github.com/dotnetcore/CAP/blob/0d8e9357d26a57140528141b30feb926e6a53bdb/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs#L94)event is not executed in the main thread, but in a background service .

When using Propagator.Inject automatically the IsRemote property is considered true.

Impact:
When using Dynatrace OneAgent we encountered problems when logging the Activity referring to the publish message, because without the variable IsRemote = true dynatrace is unable to create the necessary links.

yang-xiaodong:
Hi,

I did not clearly understand what you meant; isn't it Propagator.Inject in CapEvents.BeforePublish? If you are willing to submit a PR, I will review it.

yang-xiaodong:
No Response, Closed!

phmonte:
Sorry for the delay, I opened the [PR](https://github.com/dotnetcore/CAP/pull/1402)

PR #1402, opened 2023-09-12
fix: OpenTelemetry Dynatrace - IsRemote flag

phmonte:
### Description:
When using the _contexts dictionary to convey the context, the [IsRemote ](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitycontext.isremote?view=net-8.0#system-diagnostics-activitycontext-isremote) flag is automatically considered false, which is why [dynatrace ](https://www.dynatrace.com/) using opentelemetry cannot bind the events correctly.

According to Microsoft documentation, the [IsRemote](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitycontext.isremote?view=net-8.0#system-diagnostics-activitycontext-isremote) property indicates that the activity context was propagated from a remote parent.

As the message is sent to RabbitMQ by a background service, we need to create this link through propagation.

#### Issue(s) addressed:
- [1387](https://github.com/dotnetcore/CAP/issues/1387)

#### Changes:
- Adding compatibility with Dynatrace, without breaking other monitoring tools.

#### Affected components:
- OpenTelemetry

#### How to test:
I used dynatrace to test the success of the change and Zipkin to ensure it didn't cause any problems.

### Additional notes:
Dynatrace before changes:
image

Dynatrace after changes:
image

Zipkin after changes:
image

Zipkin before changes: 
image

The real difference is using Propagator.Inject for the IsRemote flag to be set to true, without further changes.

### Checklist:
- [N/A] I have added necessary documentation (if applicable)
- [N/A] I have updated the relevant tests (if applicable)
- [ x] My changes follow the project's code style guidelines

### Reviewers:
- @yang-xiaodong

### DIFF
--- a/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs
+++ b/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs
@@ -58,7 +58,13 @@ internal class DiagnosticListener : IObserver<KeyValuePair<string, object?>>
                     activity.AddEvent(new ActivityEvent("CAP message persistence start...",
                         DateTimeOffset.FromUnixTimeMilliseconds(eventData.OperationTimestamp!.Value)));
 
-                    if (parentContext != default) _contexts[eventData.Message.GetId()] = Activity.Current!.Context;
+                    if (parentContext != default)
+                    {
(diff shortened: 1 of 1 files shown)


yang-xiaodong (approved):
Looks good, thanks for your PR!

PR #1407, opened 2023-10-09
Fix parent span for "Event Persistence" activity trace

blashbul:
### Description:
"Event Persistence" span doesn't get context if a parent transaction is running (for example, an http request handled by a controller).
Since .Net 7, OpenTelemetry activity source name have changed to "Microsoft.AspNetCore"
cf https://github.com/open-telemetry/opentelemetry-dotnet/blob/3c2bb7c93dd2e697636479a1882f49bb0c4a362e/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L48

To a larger extent, "Event Persistence" span should be linked to a parent activity if exisit

#### Affected components:
- DotNetCore.CAP.OpenTelemetry


yang-xiaodong (approved):
Looks good, Thanks for your PR

PR #1409, opened 2023-10-16
Be able to configure if to subscribe to custom producer topic

demorgi:
### Description:
For azure service bus - `.ConfigureCustomProducer` does not imply it would create a new subscription, this PR is intended to introduce a `.WithSubscription` method to explicitly state that a subscription is needed while still maintaining backward compatibility.
Problem with subscription - if there are no suited consumers - CAP with throw exception, consider a scenario where a service consumes from `topic-a` and posts to `topic-b` via `ConfigureCustomProducer`. There are no expectations where service will consume events from `topic-b` so creating subscription is not needed.

#### Changes:
-  Updated `ServiceBusProducerDescriptorBuilder.cs` by introducing `UseTopic` and `WithSubscription` methods while maintaining backward compatibility for `ConfigureCustomProducer` method

#### Affected components:
- `ServiceBusProducerDescriptorBuilder.cs`
-  `AzureServiceBusConsumerClient.cs`

#### How to test:
Run `ServiceBusProducerBuilderTests.cs`

### Checklist:

### Reviewers:
- @mviegas
- @yang-xiaodong

### DIFF
--- a/samples/Sample.AzureServiceBus.InMemory/Program.cs
+++ b/samples/Sample.AzureServiceBus.InMemory/Program.cs
@@ -33,8 +33,8 @@ builder.Services.AddCap(c =>
             new("IsFromSampleProjectFilter","IsFromSampleProject = 'true'")
         };
 
-        asb.ConfigureCustomProducer<EntityCreatedForIntegration>(cfg => cfg.WithTopic("entity-created"));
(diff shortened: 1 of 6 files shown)


demorgi:
Hey @mviegas, @yang-xiaodong, any updates on this pull request?

yang-xiaodong (approved):

Issue #1432, opened 2023-11-16
Open telemetry subscriber thows null reference when using azure service bus without connection string


demorgi:
Open telemetry diagnostic listener fails when Azure service bus is used without a connection string.
Azure service bus supports automatic connections when using managed identities, so you don't have to specify connection string. Diagnostic listener has this code that throws a null reference as `BrokerAddress.Endpoint` is null.
https://github.com/dotnetcore/CAP/blob/7ba52a8defa73f3866001dea6dd6af434574c3d2/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs#L108C1-L109C1

Workaround #1
Explicitly set the connection string for Azure service bus to string empty when registering CAP

yang-xiaodong:
Hi @demorgi ,

Thanks for your report, you're right , would you want to submit a PR to fix it?

yang-xiaodong:
Fixed in version 8.0.0

PR #1433, opened 2023-11-18
Open telemetry subscriber thows null reference when using azure service bus without connection string

demorgi:
### Description:
Open telemetry diagnostic listener fails when Azure service bus is used without a connection string.
Azure service bus supports automatic connections when using managed identities, so you don't have to specify connection string. Diagnostic listener has this code that throws a null reference as BrokerAddress.Endpoint is null.

#### Issue(s) addressed:
- https://github.com/dotnetcore/CAP/issues/1432

#### Changes:
- BrokerAddress.Endpoint is never null, fallbacks to String.Empty

#### Affected components:
- BrokerAddress.cs

### Checklist:

### Reviewers:
- @yang-xiaodong


yang-xiaodong (approved):
LGTM, Thanks

Issue #1439, opened 2023-11-23
Outbox pattern does not work when using DotNetCore.CAP.InMemoryStorage


davidterins:
1. Create a subscriber and a publisher using DotNetCore.CAP.NATS and DotNetCore.CAP.InMemoryStorage where the publisher sends messages with an interval
2. Run the application and the nats-server
3. Shutdown nats-server
4. Keep the server shut down while some messages are being sent by the publisher.
5. Start nats-server

Expected behavior here is that the messages sent by the publisher during shutdown of the server should be received by the consumer when nats-server i started again, but they will get stuck in the publishers outbox due to exceptions since the Origin property is not set on the stored messages when retrieved from the outbox.

davte-beijer:
Thank you, when can I expect this to be released @yang-xiaodong?

yang-xiaodong:
Hi @davte-beijer ,

Fixed in version `7.2.3-preview-217562936`

yang-xiaodong:
Fixed in version 8.0.0

PR #1440, opened 2023-11-23
Set Origin property when creating new MemoryMessage for DotNetCore.CAP.InMemoryStorage

davidterins:
### Description:
Outbox pattern did not work as expected when using DotNetCore.CAP.InMemoryStorage, since Origin was null causing exceptions later on.

E.g. when method GetPublishedMessagesOfNeedRetry() in class InMemoryStorage returns the base class MediumMessage the required Origin property will not be set as compared to storage providers like SQL. The IDataStorage.SqlServer GetPublishedMessagesOfNeedRetry() returns MediumMessages with the Origin property set based the a deserialized Content property value. However the serialization/deserialization part can be ignored for InMemoryStorage.cs, since it will not be written to a database, therefore the Origin property can be be set immediately when storing the message instead in StoreMessageAsync().

#### Issue(s) addressed:
 #1439 

#### Changes:
- Set Origin property of Memory message in StoreMessageAsync()

#### Affected components:
- DotNetCore.CAP.InMemoryStorage
   -  IDataStorage.InMemory

#### How to test:
Reproduction steps in described in #1439

### Checklist:
- [ ] I have added necessary documentation - Not applicable
- [ ] I have updated the relevant tests - Not applicable

### DIFF
--- a/src/DotNetCore.CAP.InMemoryStorage/IDataStorage.InMemory.cs
+++ b/src/DotNetCore.CAP.InMemoryStorage/IDataStorage.InMemory.cs
@@ -90,6 +90,7 @@ internal class InMemoryStorage : IDataStorage
         {
             DbId = message.DbId,
             Name = name,
+            Origin = message.Origin,
             Content = message.Content,
             Retries = message.Retries,
             Added = message.Added,


yang-xiaodong (approved):
Looks Good, Thanks for your PR!

Issue #1449, opened 2023-12-07
DotNetCore.Cap.NATS do not handle reconnect if the nats server is forcibly shutdown and then restarted


davidterins:
# Issue: DotNetCore.Cap.NATS do not handle reconnect if the nats server is forcibly shutdown and then restarted

## Reproduction steps
1. Create a subscriber and a publisher using DotNetCore.CAP.NATS and DotNetCore.CAP.InMemoryStorage where the publisher sends messages to a topic with an interval
2. Run the application and the nats-server
3. Shut down the nats-server.exe by killing the process... e.g. via windows via TaskManager
4. Keep the server shut down while some messages are being sent by the publisher.
5. Start nats-server

Expected behavior here is that the messages sent by the publisher during shutdown of the server should be received by the consumer when nats-server i started again, but they will get stuck in the publishers outbox **since the subscriber will not be able to reconnect back on to the serve again**

## Suggested change
The issue is in the following method `DisconnectedEventHandler` in `NATSConsumerClient.cs` handling the disconnected events of the nats server.

### Current implementation 
```csharp
        private void ClosedEventHandler(object? sender, ConnEventArgs e)
        {
            if (e.Error is { Message: "Server closed the connection." })
            {
                var logArgs = new LogMessageEventArgs
                {
                    LogType = MqLogType.ConnectError,
                    Reason = e.Error.ToString()
                };
                OnLogCallback!(logArgs);
            }
        }
```
### New implementation

```csharp
        private void ClosedEventHandler(object? sender, ConnEventArgs e)
        {
            if (e.Error is null) return
            
            var logArgs = new LogMessageEventArgs
            {
                 LogType = MqLogType.ConnectError,
                 Reason = e.Error.ToString()
             };
             OnLogCallback!(logArgs);
        }
```

I have noticed that this ClosedEventHandler is triggered in a lot of cases where the `e.Error` is `null` and I assume that is why the explicit message check was set to ` if (e.Error is { Message: "Server closed the connection." })` in commit c63c2891, however this message do not seem to cover all the server disconnect cases. Therefore I suggest to just return when `e.Error` is `null` rather than adding another explicing message check for _"Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.."_, which is sent when nats server is shutdown forcibly.

yang-xiaodong:
[...]

PR #1450, opened 2023-12-07
Removed explicit message check with null check.

davidterins:
See description in related issue #1449

#### Changes:
- Removed explicit message check with null check, since crucial error message was missed.

#### Affected components:
- DotNetCore.Cap.NATS
   - Specifically consumers

#### How to test:
See related issue

### Checklist:
- [ ] I have added necessary documentation (if applicable) - **(Not applicable)**
- [ ] I have updated the relevant tests (if applicable) - **(Not applicable)**


yang-xiaodong (approved):
Not sure if Error to null will work in all cases, but it looks good

Issue #1452, opened 2023-12-07
OpenTelemetry: Context lost on consumer retry and Baggage Propagation


bschwehn:
Hi,

We had an issue that OpenTelemetry context propagation to the consumer is not working correctly on retry.

The issue seems to be, that the context is extracted in BeforeConsume:

https://github.com/dotnetcore/CAP/blob/67c882a6747ae36372f5081c272110f0988f12a5/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs#L149

and then saved:

https://github.com/dotnetcore/CAP/blob/67c882a6747ae36372f5081c272110f0988f12a5/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs#L172

Then the first BeforeSubscribeInvoke removes the saved context:

https://github.com/dotnetcore/CAP/blob/67c882a6747ae36372f5081c272110f0988f12a5/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs#L205

On retries, BeforeConsume is not called again and the saved context is no longer saved, thus causing the issue.

I have tried changing the code to not save the context and instead just extract it every time in BeforeSubscribeInvoke, and that solves our issue.
But I don't know why the context is saved in the first place -- is it just a performance improvement?

I can create a pull request if you think this way of fixing the issue is correct.

I have also changed the Propagator from TraceContextPropagator

https://github.com/dotnetcore/CAP/blob/67c882a6747ae36372f5081c272110f0988f12a5/src/DotNetCore.CAP.OpenTelemetry/DiagnosticListener.cs#L26

to Propagators.DefaultTextMapPropagator and added a

Baggage.Current = propagatedContext.Baggage;

which then automatically propagates Baggage.Current, which may also be useful for other people?

Instead or Propagators.DefaultTextMapPropagator, it can explicitly be set to

    private static readonly TextMapPropagator Propagator = new CompositeTextMapPropagator(
        new TextMapPropagator[]
        {
            new TraceContextPropagator(),
            new BaggagePropagator()
        });

for BaggagePropagation, but as I understand, Propagators.DefaultTextMapPropagator will allow users of CAP to enable or disable baggage propagation by calling  Sdk.SetDefaultTextMapPropagator (https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/README.md?plain=1#L463)

bschwehn:
Thanks for the quick reply.
[...]

Issue #1100, opened 2022-03-08
Question about Opentelemetry


MahmoudSamir101:
Hello there,
I saw a behavior that I'm not sure was intended, but allow me to explain.

We had a web app that made a call to a gateway, which accepted the request and published it to a queue using CAP.
As a worker, we had a service that subscribed to that queue as well via CAP.

After a successful subscription invocation by CAP, the worker publish event occurs.
Everything is great, and diagnostics from web , gateway and worker are exported to Jaeger, including CAP diagnostics, and activities are linked to each other with the exception of the last step in the flow, which is publishing event following subscription invocation.

That phase is exported as a distinct activity without parent context.

When I checked source code, I saw that there is a condition that states that if the current activity source does not belong to Aspnetcore Instrumentation, current activity will be null.

To summarize, messages published via Aspnetcore requests on the gateway are exported as child activities to web app main activity , but messages published via subscription invoking are exported as new activities, even if they are belong to the web app main activity.

So, my question is, is this intended by design ?

yang-xiaodong:
Hi @MahmoudSamir101 ,

Yes, in the implementation process to reduce complexity, a new activity is created when in the subscriber publishes message.

However, I think this is a feature that could be improved, would you like to submit a PR for this?

MahmoudSamir101:
I would like to, I will be able to work on it after two weeks.

If you have design considerations, or any thoughts share it please.

yang-xiaodong:
Great , Thanks

PR #1454, opened 2023-12-08
Fixes for OpenTelemetry Context Propagation

bschwehn:
Fixes an issue that caused context propagation to fail on message retry (#1452)
Before, BeforeSubscribeInvoke removed the context so it was no longer available on retry.

It also removes the _contexts dictionary, as during Subscribe this handling didn't work during retry and during Publish, the dictionary value was no longer used.

I have tested this with both aspnetcore services and console subscriber that publishes during the subscribe and context propagation was working fine. Thus #1100 is also closed, probably already in #1407.

#### Issue(s) addressed:
- #1452 
- #1100

#### Changes:
1. Fixes an issue that caused context propagation to fail on message retry (#1452)
Before, BeforeSubscribeInvoke removed the context so it was no longer available on retry.
2. Enable Baggage propagation
2. removes _context dictionary

#### Affected components:
- dotnetcore.CAP.OpenTelemetry

#### How to test:
- In an OpenTelemetry enabled project, publish a message. Note that the message has traceparent and baggage fields set.
- In a subscriber receiving a message with traceparent and baggage field set, note that the trace context is restored and Baggage.Current set.

While the change to fix the context propagation is a pure bugfix (imo), the change to add Baggage propagation is a (non-breaking) change. I have added to the documentation how to opt out of Baggage propagation for those that do not want to have this feature. Hopefully this change in non-controversial because
a) baggage propagationis generally the expected behaviour when DefaultTextMapPropagator includes BaggagePropagator. It is only because this is the default that we may have some people that do not actually want this but still have BaggagaPropagator
b) it is possible to disable BaggagePropagator using the standard OpenTelemetry SDK functionality
c) without this change, there was no way for people to enable BaggagePropagation in CAP at all.

### Checklist:

### Reviewers:
- @yang-xiaodong

### DIFF
--- a/docs/content/user-guide/en/monitoring/opentelemetry.md
+++ b/docs/content/user-guide/en/monitoring/opentelemetry.md
@@ -26,6 +26,40 @@ services.AddOpenTelemetryTracing((builder) => builder
 );
 ```
 
+If you don't use a framework that does this automatically for you (like aspnetcore), make sure you enable a listener, for example:
+
+```C#
+ActivitySource.AddActivityListener(new ActivityListener()
+{
+    ShouldListenTo = _ => true,
+    Sample = (ref ActivityCreationOptions<ActivityContext> _) => ActivitySamplingResult.AllData,
+    ActivityStarted = activity => Console.WriteLine($"{activity.ParentId}:{activity.Id} - Start"),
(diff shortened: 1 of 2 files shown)


yang-xiaodong (approved):
Awsome, thanks for the great work

PR #1455, opened 2023-12-11
- RetryFailedMessage cooldown time is now dynamic and configurable.

apatozi:
### Description:
For failed messages after the first 3 retries, there was a cooldown time of **4 minutes** that could not be changed. This commit request has it as an option that you can set when configuring CAP. 

#### Issue(s) addressed:
- Have not opened an issue because I solved it directly 

#### Changes:
- Added Option RetryCoolDownSeconds 
- Added argument coolDownTime of type Timespan in GetPublishedMessagesOfNeedRetry
- Added argument coolDownTime of type Timespan in GetReceivedMessagesOfNeedRetry
- Changed IProcessor.NeedRetry to call above functions using the RetryCoolDownSeconds as argument . 

#### Affected components:
- DotNetCore.Cap
- DotNetCore.Cap.InMemoryStorage
- DotNetCore.Cap.MongoDB
- DotNetCore.Cap.MySql
- DotNetCore.Cap.PostgreSql
- DotNetCore.Cap.SqlServer

#### How to test:
- Set RetryCoolDownSeconds option other than 240 seconds (4 minutes) when initializing cap 
- Setup a consumer that throws an exception 
- Measure time of the 4-th retry

- As I can't speak Chinese I cannot update the documentation to include the change. I only did it for the english documentation.

### Checklist:

### Reviewers:
- any

### DIFF
--- a/docs/content/user-guide/en/cap/configuration.md
+++ b/docs/content/user-guide/en/cap/configuration.md
@@ -82,7 +82,7 @@ During the message sending process if message transport fails, CAP will try to s
 During the message sending process if consumption method fails, CAP will try to execute the method again. This configuration option is used to configure the interval between each retry.
 
 !!! WARNING "Retry & Interval"
-    By default if failure occurs on send or consume, retry will start after **4 minutes** (RetryCoolDownSeconds) in order to avoid possible problems caused by setting message state delays.    
+    By default if failure occurs on send or consume, retry will start after **4 minutes** (FallbackWindowLookbackSeconds) in order to avoid possible problems caused by setting message state delays.    
     Failures in the process of sending and consuming messages will be retried 3 times immediately, and will be retried polling after 3 times, at which point the FailedRetryInterval configuration will take effect.
 
 !!! WARNING "Multi-instance concurrent retries"
@@ -113,7 +113,7 @@ Number of consumer threads, when this value is greater than 1, the order of mess
 Maximum number of retries. When this value is reached, retry will stop and the maximum number of retries will be modified by setting this parameter.
 
 
-#### RetryCoolDownSeconds
+#### FallbackWindowLookbackSeconds
 
 > Default: 240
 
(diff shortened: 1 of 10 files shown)


yang-xiaodong:
The 4 minutes is to ensure that newly added messages are not picked up by the retrieval process. If the user does not have a deep understanding of this configuration and sets a value that is too small, repeated message problems will easily occur. What benefits do you think settings provide as configuration items? Do you have a use case where you send a large number of messages in a short period of time?

apatozi:
Yes i have a business rule that I need to time out the entire saga if the payment doesn't process in 2 minutes. With the current rules I have to fail it if 3 retires fail because the next retry will be 4 minutes later. Usually after 30 seconds is more than enough but at most a minute works for my case. The third party payment service comes back in a shorter time than 1 minutes always so i have to either change this setting for that microservice or implement a more complex mechanism to force retry and i would like to have this included in the library. For the moment i am using a custom build of the library with this setting set but maybe some other people want the feature also so i decided to make a PR.

yang-xiaodong:
I think you should not use this feature to implement payment timeout retries but should use some way that conforms to business rules.

If you set the time too short, such as 30 seconds, if the publisher sends a large number of messages at once, and these messages cannot be processed within 30 seconds, they will be picked up by the retry processor and cause repeated execution.

yang-xiaodong:
I think customizing this value makes sense in some use cases, but we need to use appropriate methods to remind users when setting a value that is too small, such as printing an alarm log if it is detected that it is less than 30 seconds.

apatozi:
Yes a warning would be a wise thing to do. I could help with that if you want and commit it. Let me know.

yang-xiaodong:
Yes, The option name should be changing, how about  `FallbackWindowLookbackSeconds`?

yang-xiaodong (approved):
Looks good, Thanks for your PR!

PR #1456, opened 2023-12-12
Fix: message retried forever SubscriberNotFound

bschwehn:
If we have a message in the datastore pending retry, and we no longer have the subscriber, the message is never cleared from the data store and retried forever.

This can happen during software updates for example (Subscriber removed but pending messages are still in the data store).

Since the message was not updated in the datastore, the message would be retried forever.

Fix by calling SetFailedState which will update the msg in the datastore.

I think this handling was similar in the past and was lost in some refactoring. SetFailedState still has the handling to set retries to the max retries, so that such messages are not retried either.

#### Changes:
- Fix infinite retry of messages after subscriber is removed.

#### Affected components:
- Subscriber / Retry handling

#### How to test:
- publish a message that has a subscriber, fail in the subscriber so that the message is put into retry queue
- remove subscriber
- observe that message is retried forever

### Checklist:


yang-xiaodong (approved):
Looks good, Thanks

PR #1469, opened 2024-01-14
Fixed infinite loop problem in IConnectionChannelPool.Rent() method when its disposed

yunusemrecebe:
### Description:
When using the IConnectionChannelPool.Rent() method, when a large number of operations are performed with parallel threads, a message is canceled immediately after being published, but the IConnectionChannelPool.Rent method cannot process this information yet. The value of the _maxSize variable becomes 0 and the loop becomes infinite.

#### Issue(s) addressed:
- None

#### Changes:
- IConnectionChannelPool.Rent() method updated and new property inserted to ConnectionChannelPool class with named liked _isDisposed

#### Affected components:
- src/DotNetCore.CAP.RabbitMQ/IConnectionChannelPool.Default.cs

#### How to test:
_You should create a minimum of 1000 different parallel threads at the same time and constantly sends intense requests. Cancel the requested cancellation times after some time. Then the _maxSize variable of the intensive processor processor will be set to 0 and the infinite loop will occur.

### Checklist:
- [x ] I have tested my changes locally
- [ x] My changes follow the project's code style guidelines

### Reviewers:
- Sorry I don't know anyone :/

### DIFF
--- a/src/DotNetCore.CAP.RabbitMQ/IConnectionChannelPool.Default.cs
+++ b/src/DotNetCore.CAP.RabbitMQ/IConnectionChannelPool.Default.cs
@@ -22,6 +22,7 @@ public class ConnectionChannelPool : IConnectionChannelPool, IDisposable
     private readonly ILogger<ConnectionChannelPool> _logger;
     private readonly ConcurrentQueue<IModel> _pool;
     private IConnection? _connection;
+    private bool _isDisposed;
 
     private int _count;
     private int _maxSize;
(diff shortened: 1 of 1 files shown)


yang-xiaodong:
Thanks for your feedback,

In the past and just now, I used `wrk` and `apache ab` to do benchmark tests, and I cancelled them at any time. I didn't encounter this issue.

Just now, I also directly created a loop and still couldn't reproduce it. The code is as follows:

```cs
for (int i = 0; i < 1500; i++)
{
   _ = Task.Run(async () =>
   {
       CancellationTokenSource cts = new CancellationTokenSource();
       if (Random.Shared.Next(1, 5) == 5)
       {
           cts.CancelAfter(Random.Shared.Next(1, 15));
       }
       await _capBus.PublishAsync("sample.rabbitmq.test", DateTime.Now, cancellationToken: cts.Token);
   });
}
```
`Dispose()` method is only called in two places: when DI is released or in the `GetConnection()`  method, but there is an lock there,  Can you show me your test code?

PR #1478, opened 2024-01-26
Respect the node port while configuring downstream request (GatewayProxyAgent)

3ldar:
### Description:
In the k8n discovery mode service address is split into two parts one is the HTTP address and the other one is the port. So the port is never included in the service address. `GatewayProxyAgent` assumes that if an address starts with `http` it must be a full address and ignores the port and this causes a request to a nonexistent address. This PR fixes the issue by respecting the port info that comes from the node discovery.

#### Affected components:
- Dashboard

#### How to test:
- Use k8n node discovery in a cluster where services run on different ports than 80 and switch from dashboard.

### Checklist:
- [x] I have added the necessary documentation (if applicable)

### Reviewers:
- @yang-xiaodong


yang-xiaodong (approved):
LGTM, Thanks for your PR!

Issue #435, opened 2019-11-20
Support custom table names


lhysrc:
因为表名中包含点号(.)，在内部规范中不允许。能否提供表名自定义，或者至少分隔符也添加到配置（CapOptions）中。

yang-xiaodong:
Hello, we will improve this feature in the next version, thank you for your feedback.

usama288:
when are you rolling out this request ?

yang-xiaodong:
In the v3.0 version, you can inherit the StorageInitializer and override the related method of getting the table name to achieve this.

MySql Example: 
```C#
public class MyCapTableNameInit : MySqlStorageInitializer
{
    public MyCapTableNameInit(ILogger<MySqlStorageInitializer> logger, IOptions<MySqlOptions> options) : base(logger, options)
    {
    }

    public override string GetPublishedTableName()
    {
        // Your message published table name
    }

    public override string GetReceivedTableName()
    {
        // Your message received table name
    }
}
```

Inject the custom implementation :

```C#
services.AddSingleton<IStorageInitializer, MyCapTableNameInit>();
```

PR #1490, opened 2024-02-22
fix(#435): use TryAddSingleton when registering IStorageInitializer to avoid overwriting user configured initialisers

esskar:
### Description:
It was not obvious how this feature works. adding a custom IStorageInitializer before calling UseCap is not working as the registration will get overwritten. This is not the case for the MySQL Initializer as this was already using the TryAddSingleton functionality

#### Issue(s) addressed:
- #435 

#### Changes:
- Calling TryAddSingleton when registering IStorageInitializer

#### Affected components:
- InMemoryCapOptionsExtensions
- MongoDBCapOptionsExtensions
- PostgreSqlCapOptionsExtensions
- SqlServerCapOptionsExtensions

I noticed that setting a custom IStorageInitializer

### Checklist:


demorgi:
as a temporary workaround you can override registration
After 
`options.UseMongoDB`
do 
`
        services.RemoveAll<IStorageInitializer>();
        services.AddSingleton<IStorageInitializer, CustomStorageInitializer>();
`

Issue #1310, opened 2023-04-11
Support PostgreSQL using ProvidePasswordCallback for azure ad


mikara89:
I am using option.UseEntityFramework(...) with Postgres and 

```
 services.AddCap(options =>
            {
                options.UseEntityFramework<CapDbContext>(opt => opt.Schema = "Cap");
                options.UseInMemoryMessageQueue();
                options.UseDashboard(options =>
                {
                    options.UseAuth = true;
                    options.DefaultAuthenticationScheme = "...";
                });

            });
```
```
private static IServiceCollection AddPGSQL<T>(this IServiceCollection services, string connectionString) 
            where T : DbContext
        {
            services.AddDbContextFactory<T>(m => m.UseNpgsql(connectionString, e => {
                    e.ProvidePasswordCallback((h, p, d, u) => {
                        var tokenCredential = new DefaultAzureCredential();
                        var AccessToken = tokenCredential.GetToken(new Azure.Core.TokenRequestContext(
                            new[] { "https://ossrdbms-aad.database.windows.net/.default" }
                            ));

                        return AccessToken.Token;
                    });
                    e.MigrationsAssembly(typeof(T).Assembly.FullName);
                }), ServiceLifetime.Scoped);
            return services;
        }
```
for using access token as password on db but cap is not triggering `ProvidePasswordCallback` at all.
Is this bug or what?

yang-xiaodong:
I was not aware of ProvidePasswordCallback before. We obtain the connection string from DbContext for ADO.NET database connections, and we do not dynamically update the connection string during its use.
https://github.com/dotnetcore/CAP/blob/191937eb799e7c847abfd7b36fbfdad55a2cf8d1/src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs#L42

If you are interested add this feature, welcome submit a PR.

mikara89:
> I was not aware of ProvidePasswordCallback before. We obtain the connection string from DbContext for ADO.NET database connections, and we do not dynamically update the connection string during its use.
> 
> https://github.com/dotnetcore/CAP/blob/191937eb799e7c847abfd7b36fbfdad55a2cf8d1/src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs#L42
> 
> If you are interested add this feature, welcome submit a PR.

[...]

PR #1496, opened 2024-03-05
Allow configuring PostgreSQL using an Npgsql data source

jonekdahl:
### Description:
Add an optional DataSource property in the PostgreSqlOptions. Using a data source will allow users of CAP to configure the full range of Npgsql options without having to expose each one in the CAP API. In particular, this allows to configure [dynamic password rotation](https://www.npgsql.org/doc/security.html#auth-token-rotation-and-dynamic-password) needed when authenticating against certain databases hosted in the cloud.

#### Issue(s) addressed:
- #1310

#### Changes:
- Allow configuration of `NpgsqlDataSource` through a new `DataSource` property in `PostgreSqlOptions`.
- Create `NpgsqlConnection` from data source if there is one, otherwise fall back to using the connection string

#### Affected components:
- PostgresSQL

#### How to test:
The new data source configuration can be tested like this:

```csharp
        var dataSource = new NpgsqlDataSourceBuilder(connectionString).Build();

        services.AddCap(capOptions =>
            capOptions
                // other CAP config
                .UsePostgreSql(opts => opts.DataSource = dataSource);
            )
```

My motivation for creating this PR was to allow CAP to be used with a PostgreSQL database hosted in Azure and authenticated via Azure managed identity (this requires configuring a password provider on the data source since the password tokens expire). In order to test this specific setup you need an Azure account with Entra ID and an Azure PostgreSQL server that has been configured to [sign in with Entra](https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/how-to-configure-sign-in-azure-ad-authentication).

An example of how to use an Azure credential to a get an access token to use as a database password is provided by Microsoft [here](https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/how-to-connect-with-managed-identity#connect-using-managed-identity-in-c).

The remaining piece is to configure the data source with a password provider that refreshes that token:

```csharp
        dataSourceBuilder.UsePeriodicPasswordProvider(...); // password callback fetching a new access token from Azure
```

`PostgreSqlOptions.ConnectionString` has been kept for backwards compatibility, removing it would be a breaking change. In the future, it might not be necessary to have both DataSource and ConnectionString in PostgreSqlOptions.

The [data source concept](https://www.npgsql.org/doc/basic-usage.html#data-source) was introduced in Npgsql 7.0 and is now the recommended way to create connections. Direct instantiation of a NpgsqlConnection is still supported, but [discouraged](https://www.npgsql.org/doc/basic-usage.html#connections-without-a-data-source).

### Checklist:

### Reviewers:
@yang-xiaodong

### DIFF
--- a/src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
+++ b/src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
@@ -16,6 +16,7 @@ public class PostgreSqlOptions : EFOptions
     /// <summary>
     /// Gets or sets the database's connection string that will be used to store database entities.
     /// </summary>
+    [Obsolete("Use .DataSource = NpgsqlDataSource.Create(<connectionString>) for same behavior.")]
     public string ConnectionString { get; set; } = default!;
     
     /// <summary>
--- a/src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
+++ b/src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
@@ -6,6 +6,7 @@ using DotNetCore.CAP.Internal;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
+using Npgsql;
 
 // ReSharper disable once CheckNamespace
 namespace DotNetCore.CAP;
@@ -16,6 +17,19 @@ public class PostgreSqlOptions : EFOptions
     /// Gets or sets the database's connection string that will be used to store database entities.
     /// </summary>
     public string ConnectionString { get; set; } = default!;
+    
+    /// <summary>
+    /// Gets or sets the Npgsql data source that will be used to store database entities.
+    /// </summary>
+    public NpgsqlDataSource? DataSource { get; set; }
+
+    /// <summary>
+    /// Creates an Npgsql connection from the configured data source.
+    /// </summary>
+    public NpgsqlConnection CreateConnection()
+    {
+        return DataSource != null ? DataSource.CreateConnection() : new NpgsqlConnection(ConnectionString);
+    }
 }
 
 internal class ConfigurePostgreSqlOptions : IConfigureOptions<PostgreSqlOptions>
--- a/src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
+++ b/src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
@@ -26,7 +26,7 @@ public class PostgreSqlOptions : EFOptions
     /// <summary>
     /// Creates an Npgsql connection from the configured data source.
     /// </summary>
-    public NpgsqlConnection CreateConnection()
+    internal NpgsqlConnection CreateConnection()
     {
         return DataSource != null ? DataSource.CreateConnection() : new NpgsqlConnection(ConnectionString);
     }
--- a/docs/content/user-guide/en/storage/postgresql.md
+++ b/docs/content/user-guide/en/storage/postgresql.md
@@ -32,10 +32,11 @@ public void ConfigureServices(IServiceCollection services)
 
 #### PostgreSqlOptions
 
-NAME | DESCRIPTION | TYPE | DEFAULT
-:---|:---|---|:---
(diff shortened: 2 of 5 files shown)


src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
```diff
@@ -16,6 +17,19 @@ public class PostgreSqlOptions : EFOptions
...
+    /// </summary>
+    public NpgsqlDataSource? DataSource { get; set; }
+
+    /// <summary>
+    /// Creates an Npgsql connection from the configured data source.
+    /// </summary>
+    public NpgsqlConnection CreateConnection()
+    {
+        return DataSource != null ? DataSource.CreateConnection() : new NpgsqlConnection(ConnectionString);
+    }
```
jonekdahl, line 32: Is it okay to have this helper method in `PostgreSqlOptions` or should it be moved somewhere else?

README.md
```diff
@@ -70,13 +70,17 @@ public void ConfigureServices(IServiceCollection services)
...
+        x.UseEntityFramework<AppDbContext>(); //Options, Notice: Normally you don't need to configure x.UseSqlServer("") again! CAP can pick up the connection string from the EF context.
 
         // If you are using ADO.NET, choose to add configuration you needed：
         x.UseSqlServer("Your ConnectionStrings");
         x.UseMySql("Your ConnectionStrings");
         x.UsePostgreSql("Your ConnectionStrings");
 
+        // If you are using an Npgsql data source (to set password provider etc.), tell CAP to use it like this:
+        var dataSource = new NpgsqlDataSourceBuilder("Your ConnectionStrings").Build();
+        x.UsePostgreSql(options => options.DataSource = dataSource);
```
jonekdahl, line 82: I added an example here to show how it would be used. Let me know if you think it's too much detail to include here.

yang-xiaodong: I think the special instructions for postgresql can be explained in our docs and keep the readme concise.

Docs here:
https://github.com/dotnetcore/CAP/blob/master/docs/content/user-guide/en/storage/postgresql.md

src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
yang-xiaodong: Here is ok, Should we change the accessibility from `public` to `internal`?

yang-xiaodong (changes requested):

README.md
jonekdahl: Okay, reverted this and added the property in the the storage docs instead.

src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
jonekdahl: Good suggestion. Done.

src/DotNetCore.CAP.PostgreSql/CAP.PostgreSqlOptions.cs
```diff
@@ -16,6 +17,19 @@ public class PostgreSqlOptions : EFOptions
     /// Gets or sets the database's connection string that will be used to store database entities.
     /// </summary>
     public string ConnectionString { get; set; } = default!;
```
mviegas, line 19: Just some food for thought here: I know that we're keeping this for backward compatibility purposes, but would it make sense to mark it as obsolete and indicate that we intend to replace it with the `NpgsqlDataSource` in future versions?

mviegas: I think it looks great. @yang-xiaodong what's your opinion over this deprecation notice for this property?!

yang-xiaodong: Our next release 8.1.0 is a minor version, So I think `[Obsolete]` it is acceptable.

jonekdahl: Ok, I have marked ConnectionString obsolete.

yang-xiaodong (approved):
LGTM

jonekdahl:
@yang-xiaodong Thanks for merging! Would it be possible to get a preview release with this feature included? I would highly appreciate that.

PR #1498, opened 2024-03-06
Adding default publish headers

demorgi:
### Description:
A case when you need to add specific custom header to each message - it's not possible at the moment.
Added `DefaultPublishHeaders` dictionary that gets added to all published messages.

#### Changes:
* Added `DefaultPublishHeaders` dictionary to CAP.Options
* In `CapPublisher.Default` collection gets added to headers, existing headers are never overriden

#### Affected components:
* `ICapPublisher.Default`

### Checklist:
- [ x] I have tested my changes locally
- [ x] I have added necessary documentation (if applicable)
- [] I have updated the relevant tests (if applicable)
- [x ] My changes follow the project's code style guidelines

### Reviewers:
@yang-xiaodong
@mviegas

### DIFF
--- a/docs/content/user-guide/en/cap/configuration.md
+++ b/docs/content/user-guide/en/cap/configuration.md
@@ -144,6 +144,12 @@ The expiration time (in seconds) of the failed message. When the message is sent
 
 If `true` then all consumers within the same group pushes received messages to own dispatching pipeline channel. Each channel has set thread count to `ConsumerThreadCount` value.
 
+#### DefaultPublishHeaders
+
+> Default: Dictionary<string, string>.Empty
+
+When populated - automatically adds these headers to all outgoing messages, never overrides any existing default headers.
+
 #### EnableConsumerPrefetch
 
 > Default: false， Before version 7.0 the default behavior is true
--- a/src/DotNetCore.CAP/CAP.Options.cs
+++ b/src/DotNetCore.CAP/CAP.Options.cs
@@ -111,6 +111,11 @@ public class CapOptions
     public bool UseDispatchingPerGroup { get; set; }
 
     /// <summary>
+    /// Headers to be added to the message when it is published.
+    /// </summary>
+    public IDictionary<string, string> DefaultPublishHeaders { get; } = new Dictionary<string, string>();
+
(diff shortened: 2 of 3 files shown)


mviegas:
@demorgi I just remembered the following: wouldn't the current `CustomHeaders` property in `AzureServiceBusOptions` solve this issue without the need of extra code? Please check it [here](https://cap.dotnetcore.xyz/user-guide/en/transport/azure-service-bus/#heterogeneous-systems).

demorgi:
> @demorgi I just remembered the following: wouldn't the current `CustomHeaders` property in `AzureServiceBusOptions` solve this issue without the need of extra code? Please check it [here](https://cap.dotnetcore.xyz/user-guide/en/transport/azure-service-bus/#heterogeneous-systems).


Correct me if I'm wrong but CustomHeaders are for ingestion, just to support of other publishers outside of CAP (Like mass transit publishes a message, so now we need to configure CustomHeaders to injest it)

mviegas:
You're completely right, it's been a while since I touched this feature. Apologies for the confusion. 

LGTM

mviegas (approved):

yang-xiaodong:
I don't think this is a good feature. In our case, we need to add a dynamic TenantId as a header in each message. This new Option cannot achieve this. I think this requires the user to wrap Publish to add additional header information. Taking a step back, we need to provide a published AOP feature similar to SubscribeFilter?

demorgi:
@yang-xiaodong Let me explain it a little bit differently, so there are 2 ways you can get data to populate headers:
1. From request pipeline, either from request payload or fetched from datastore or etc, all of this data is a part of request processing and is disposed once request is finished. You can get data from this request and post it to headers, as this data is dynamic there is no good way to automatically populate headers (unless we start doing some context magic with lambdas expressions and stuff). In this case you can just manually add header on message publish as it needs additional control (null checks and stuff)
2. Data from environment variables, this one can be populated with this new dictionary. It can be used for message routing or provide additional host context for messages

yang-xiaodong:
@demorgi  I want to add a filter or something to support this feature in the future version, thanks for your PR!

Issue #1521, opened 2024-04-26
Incorrect behavior when event is sent outside of transaction 


NikolozGob:
We have encountered issues with using transaction and then sending an event. We were using Kafka with PostgreSQL. Issue can be replicated with [Sample project](https://github.com/dotnetcore/CAP/tree/master/samples/Sample.Kafka.PostgreSql).

Bug happens when we use database transaction and outside of transaction we send an event.

![image](https://github.com/dotnetcore/CAP/assets/164164921/0459b338-9fb2-4305-89b7-b8c9c5d00dc8)

This causes event to be added in in Published table with Scheduled status and it's sent after 4 minutes instead of immediately, even if it is a delayed event.

For example, this event will be published in 4 minutes instead of one hour:
![image](https://github.com/dotnetcore/CAP/assets/164164921/a9b70981-4b68-4b76-af6d-79d44cd487d2)

Bug cause:
Cap publisher is still holding transaction object, even though event is out of the scope of the transaction.

![image](https://github.com/dotnetcore/CAP/assets/164164921/083e0d46-b801-4702-b2ce-d658990a4efb)

yang-xiaodong:
Hi @NikolozGob ,

Thanks for your report ,  It's a bug !

The code  `if (Transaction == null)` should be `if (Transaction?.DbTransaction == null) ` ,  would you want make a PR to fix it ?

NikolozGob:
Okay, I'll make a PR

NikolozGob:
Opened pull request

yang-xiaodong:
Fixed in version 8.1.2

PR #1522, opened 2024-04-26
Fix: Transaction is nullified after finishing the scope

NikolozGob:
Fixed bug that causes CapPublisher to hold finished transactions resulting in incorrect publishing behavior

[Related issue](https://github.com/dotnetcore/CAP/issues/1521)

Changes:
Changed Transaction null check in CapPublisher

It can be tested via Sample.Kafka.PostgreSql. After Moving event publishing outside of transaction scope, event will be sent immediately or with a correct delay (if given).


yang-xiaodong (approved):
LGTM, Thanks

Issue #1536, opened 2024-05-27
Feature Request: Per-Queue Configuration for BasicQos and ConsumerThreadCount


ajdwfnhaps:
Description
Currently, it seems that RabbitMQ's BasicQos and ConsumerThreadCount parameters can only be configured globally across the entire service in CAP. It would be highly beneficial to have the ability to configure these parameters individually for each queue. This would allow for more granular control over the message processing behavior, tailored to the specific needs of different queues.

Proposed Solution
Introduce the capability to configure BasicQos and ConsumerThreadCount on a per-queue basis within CAP's configuration. This could be achieved by extending the current configuration options to accept settings for each queue.

Use Case
In scenarios where different queues have different performance and concurrency requirements, being able to specify `BasicQos

yang-xiaodong:
We plan to support consumer concurrent in v8.2.0, please check out the PR https://github.com/dotnetcore/CAP/pull/1537 !

yang-xiaodong:
@ajdwfnhaps Version  [8.2.0-preview-233720681](https://www.nuget.org/packages/DotNetCore.CAP/8.2.0-preview-233720681) is released, take a try.

ajdwfnhaps:
got it, Thanks

PR #1537, opened 2024-05-27
Add GroupConcurrent option to support subscriber concurrent execution

yang-xiaodong:
### Description:

CAP currently supports specifying the number of parallel execution threads for subscribers through the global parameter `SubscriberParallelExecuteThreadCount`. However, in some scenarios, some subscribers may need to execute serially while others may need to execute in parallel.

Subscribers need to provide a configuration to support the above behavior, while `SubscriberParallelExecuteThreadCount` still takes effect, which can be used to set the parallelism for subscribers who are not individually configured.

#### Issue(s) addressed:

- #1536

#### Changes:

1. **Removal of UseDispatchingPerGroup**

The `UseDispatchingPerGroup` configuration has been removed and is now the default behavior.

2. **Addition of GroupConcurrent parameter to `[CapSubscribe]`**

A new `GroupConcurrent` parameter has been added to the `[CapSubscribe]` attribute for concurrent execution of subscriber methods. Now, the Group corresponds to consumers on independent threads.

Usage & Example:

```cs
[CapSubscribe("sample.rabbitmq.hello", Group = "A", GroupConcurrent = 3)]
public void Subscriber(DateTime dateTime)
{
}
```

By specifying the value of the `GroupConcurrent` parameter, you set the degree of parallel execution for the subscriber, meaning that the `Subscriber` method can have up to 3 degrees of parallelism.

Parallel execution requires independent threads, so if no `Group` parameter is specified, CAP will automatically create a Group using the `Name` value.

If you have multiple subscribers set to the same Group and have also set the `GroupConcurrent` value for these subscribers, only the value set by the first subscriber will take effect.

#### Affected components:

- All Broker components.
- The Class DispatcherPerGroup has been removed.

### Checklist:

### Reviewers:
- @xiangxiren


xiangxiren (commented):
It looks very good.

Issue #1545, opened 2024-06-16
NATS feature: add the possibility to disable dynamic consumer subject/topic creation through NatsCapOtpions


davte-beijer:
Currently NATS clients will create any missing topics of they are not configured on a stream through `NATSConsumerClient.FetchTopics`. This requires extra permissions for the client to be configured in the NATS server configuration.  e.g:

```json
"permissions": {
   "publish": {
      "JS.API.STREAM.INFO.VolatileStream"
      "JS.API.STREAM.CREATE.VolatileStream.*"
      "JS.API.STREAM.UPDATE.VolatileStream.*"
   }
}
```

For use cases when you want to limit the permissions required for a nats client, by having a centralized solution for creating streams & topics, then allowing dynamic creation by clients is not feasible as it currently enforces a client to have the publish permissions specified above.

Therefore I propose to extended the `NATSOptions` with a new property `EnableSubscriberClientStreamAndTopicCreation ` with default value `true` in order for the change to be backwards compatible.

**Pseudo code:**

```csharp
public class NATSOptions
{
   ...
    /// <summary>
    /// Allows a nats client to dynamically create a stream and configure its expected topics.
    /// </summary>
    public bool EnableSubscriberClientStreamAndTopicCreation {get; set;} = true;
    ...
}
```

This option would then be used in the `NATSConsumerClient.FetchTopics` method like the following in order to avoid doing calls requiring described client publish permissions via the the JetsManagementContext class.

```csharp
public ICollection<string> FetchTopics(IEnumerable<string> topicNames)
{
    if (_natsOptions.EnableSubscriberClientStreamAndTopicCreation)
    {
        Connect();

        var jsm = _consumerClient!.CreateJetStreamManagementContext();

        var streamGroup = topicNames.GroupBy(x => _natsOptions.NormalizeStreamName(x));

        foreach (var subjectStream in streamGroup)
        {
            var builder = StreamConfiguration.Builder()
                .WithName(subjectStream.Key)
                .WithNoAck(false)
                .WithStorageType(StorageType.Memory)
                .WithSubjects(subjectStream.ToList());

            _natsOptions.StreamOptions?.Invoke(builder);

            try
            {
                jsm.GetStreamInfo(subjectStream.Key); // this throws if the stream does not exist

                jsm.UpdateStream(builder.Build());
            }
            catch (NATSJetStreamException)
            {
                try
                {
                    jsm.AddStream(builder.Build());
                }
                catch
                {
                    // ignored
[...]

PR #1556, opened 2024-06-29
Nats improvement 1545

davidterins:
### Description:
With these changes a consumer of the DotNetCore.CAP.NATS nuget package can opt out from allowing a NATS Consumer client to create topics and streams dynamically which enforces each client to be granted "admin" like permissions. To avoid assigning these permissions to each client, a consumer of the nuget package can now opt out from this by specifying NATSOptions property `EnableSubscriberClientStreamAndSubjectCreation` to `false` (defaults to true).

#### Issue(s) addressed:
- #1545 

#### Changes:
- Added new `NATSOptions` property `EnableSubscriberClientStreamAndSubjectCreation` and usage of it in `NatsConsumerClient.FetchTopics` method.
- Added call to `OnLogCallback` with `MqLogType.ConnectError` in order to mark the `ConsumerRegister` as "unhealthy" to allow "subscribe retries" calls to the `NatsConsumerClinet.Subscribe` method in case the server have not yet configured the expected stream before the client tries to subscribe.

#### Affected components:
- `DotNetCore.CAP.NATS`

#### How to test:

1. Start a nats server (with no configured streams)
2. Start application with a cap subscriber on a specific subject/topic
   Configured with NatsOptions `EnableSubscriberClientStreamAndSubjectCreation: false`
3. The client will fail to subscribe since the expected stream does not yet exist.
4. Configure the nats server with expected streams and topics expected by application in step 2
5. The application should subscribe to the now configured streams and topics.

### Checklist:
- [x] I have added necessary documentation (if applicable) **(xml summary on public api)**
- [ ] I have updated the relevant tests (if applicable) **(No existing tests for CAP)**

### Reviewers:
- @yang-xiaodong

### DIFF
--- a/src/DotNetCore.CAP.NATS/CAP.NATSOptions.cs
+++ b/src/DotNetCore.CAP.NATS/CAP.NATSOptions.cs
@@ -26,6 +26,11 @@ namespace DotNetCore.CAP
         public int ConnectionPoolSize { get; set; } = 10;
 
         /// <summary>
+        /// Allows a nats consumer client to dynamically create a stream and configure the expected subjects on the stream. Defaults to true.
+        /// </summary>
+        public bool EnableSubscriberClientStreamAndSubjectCreation { get; set; } = true;
+
+        /// <summary>
         /// Used to setup all NATs client options
         /// </summary>
         public Options? Options { get; set; }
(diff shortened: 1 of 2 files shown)


yang-xiaodong (approved):
LGTM

Issue #1571, opened 2024-08-09
Azure Service Bus Custom Consumer


KuchaD:
Current Situation:
Currently, our Azure Service Bus provider lacks support for multiple namespaces and topics for consumers. We are utilizing a single topic within Azure Service Bus for messaging, and a single namespace for the application.

Solution Attempt:
I have attempted to implement a solution that addresses this limitation, and it works as expected. However, I am currently unable to submit a pull request (PR) for this update.

Register in program files.

```
builder.Services.AddCap(c =>
{
    c.Version = "v1";
    c.UseInMemoryStorage();
    c.UseAzureServiceBus(asb =>
    {
        asb.ConnectionString =
            "Endpoint=xxx";
        asb.CustomHeadersBuilder = (message, serviceProvider) =>
        {
            var snowFlakeId = serviceProvider.GetRequiredService<ISnowflakeId>();

            return new List<KeyValuePair<string, string>>()
                        {

                            new(DotNetCore.CAP.Messages.Headers.MessageId,
                                snowFlakeId.NextId().ToString()),
                            new(DotNetCore.CAP.Messages.Headers.MessageName, message.Subject),
                            new("IsFromSampleProject", "'true'")
                        };
        };
        asb.SQLFilters = new List<KeyValuePair<string, string>>() {
            new("IsFromSampleProjectFilter","IsFromSampleProject = 'true'")
        };
        
        
        asb.ConfigureCustomProducer<EntityCreatedForIntegration>(cfg => cfg.UseTopic("entity-created").WithSubscription());
        asb.ConfigureCustomProducer<EntityDeletedForIntegration>(cfg => cfg.UseTopic("entity-deleted").WithSubscription());
        asb.ConfigureCustomConsumer(cfg =>
        {
            cfg.UseGroupName($"test.{c.Version}");
            cfg.UseTopic("entity-created");
        });
    });

    c.UseDashboard();
});
```
    
Consumer
```
    [CapSubscribe(nameof(EntityCreatedForIntegration), Group = "test")]
    public void Handle(EntityCreatedForIntegration message)
    {
        Console.WriteLine($"Message {message.Id} received");
    }
```

CustomConsumer
```
public class ServiceBusConsumerDescriptorBuilder
{
    private string GroupName { get; set; } = null!;
    private string? Namespace { get; set; } = null;
    private string TopicPath { get; set; } = null!;
    
    
    public ServiceBusConsumerDescriptorBuilder UseGroupName(string groupName)
[...]

PR #1572, opened 2024-08-09
feat: asb custom consumers

KuchaD:
### Description:
Current Situation:
Currently, our Azure Service Bus provider lacks support for multiple namespaces and topics for consumers. We are utilizing a single topic within Azure Service Bus for messaging, and a single namespace for the application.

Solution Attempt:
I have attempted to implement a solution that addresses this limitation, and it works as expected. However, I am currently unable to submit a pull request (PR) for this update.

#### Issue(s) addressed:
- https://github.com/dotnetcore/CAP/issues/1571

#### Changes:
add CustomConsumer topic

#### Affected components:
Azure service bus 

#### How to test:
_Provide a step-by-step guide on how to test your changes. Include any relevant information like environment setup, dependencies, etc._

### Checklist:
- [ ] I have tested my changes locally

### Reviewers:


mviegas:
Hi @KuchaD. I will take a deeper look into this in the following days. Meanwhile, could you also update the documentation to include the custom consumers?

docs/content/user-guide/en/transport/azure-service-bus.md
```diff
@@ -57,6 +57,7 @@ The AzureServiceBus configuration options provided directly by the CAP:
...
 | DefaultCorrelationHeaders            | Adds additional correlation properties to all [correlation filters](https://learn.microsoft.com/en-us/azure/service-bus-messaging/topic-filters#correlation-filters). | IDictionary<string, string>                          | Dictionary<string, string>.Empty |
 | SQLFilters                           | Custom SQL Filters by name and expression on Topic Subscribtion                                                                                                       | List<KeyValuePair<string, string>>                   | null                             |
+| CustomConsumers                           | Custom consumer client for group with extra specification                                                                                                       | IDictionary<string, IServiceBusConsumerDescriptor>                  | Dictionary<string, IServiceBusConsumerDescriptor>.Empty
```
mviegas, line 60: Since this property is internal, I think that we shouldn't document it here.

docs/content/user-guide/en/transport/azure-service-bus.md
```diff
@@ -109,3 +110,30 @@ c.UseAzureServiceBus(asb =>
         };
 });
 ```
+
+#### SQL Filters
```
mviegas, line 114: Just a minor glitch: shouldn't it be custom consumers instead of SQL Filters? That section makes more sense than the property since this is the API that is going to be public.

samples/Sample.AzureServiceBus.InMemory/SampleSubscriber.cs
```diff
@@ -1,14 +1,23 @@
...
     
-    [CapSubscribe("cap.sample.tests")]
+    [CapSubscribe(nameof(EntityCreated))]
     public void Handle(Message message)
     {
         Console.WriteLine($"Message {message.Content} received");
     }
+    
+    [CapSubscribe(nameof(EntityCreatedForIntegration), Group = "test")]
+    [CapSubscribe(nameof(EntityDeleted))]
```
mviegas, line 18: Are we trying to have two different subscriptions for the same message here? Just to understand the two `CapSubscribe` attributes.

src/DotNetCore.CAP.AzureServiceBus/Options/CAP.AzureServiceBusOptions.cs
```diff
@@ -0,0 +1,71 @@
...
+        configuration(builder);
+        CustomProducers.Add(builder.Build());
+
+        return this;
+    } 
+    
+    internal IDictionary<string, IServiceBusConsumerDescriptor> CustomConsumers{ get; set; } =
+        new Dictionary<string, IServiceBusConsumerDescriptor>();
+
+    public AzureServiceBusOptions ConfigureCustomGroupConsumer(string groupName,
```
mviegas, line 62: To make it more coherent with the Transport idiom, what do you think about changing it to ConfigureCustomTopicConsumer?

src/DotNetCore.CAP.AzureServiceBus/Options/AzureServiceBusOptionsBase.cs
```diff
@@ -7,42 +7,22 @@
...
-using DotNetCore.CAP.AzureServiceBus.Producer;
 
 // ReSharper disable once CheckNamespace
 namespace DotNetCore.CAP;
 
 /// <summary>
 /// Provides programmatic configuration for the CAP Azure Service Bus project.
 /// </summary>
-public class AzureServiceBusOptions
+public class AzureServiceBusOptionsBase
```
mviegas, line 17: What's the purpose of separating this in a base class?

src/DotNetCore.CAP.AzureServiceBus/AzureServiceBusConsumerClient.cs
```diff
@@ -8,16 +8,17 @@
...
+using DotNetCore.CAP.AzureServiceBus.Consumer;
 using DotNetCore.CAP.Messages;
 using DotNetCore.CAP.Transport;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace DotNetCore.CAP.AzureServiceBus;
 
-internal sealed class AzureServiceBusConsumerClient : IConsumerClient
+internal class AzureServiceBusConsumerClient : IConsumerClient
```
mviegas, line 19: Any specific reason why we're removing the sealed keyword here?

src/DotNetCore.CAP.AzureServiceBus/AzureServiceBusConsumerClient.cs
```diff
@@ -8,16 +8,17 @@
...
 {
-    private readonly AzureServiceBusOptions _asbOptions;
+    private AzureServiceBusOptions _asbOptions;
```
mviegas, line 21: Should we mutate this field here or create a new one with the CustomConsumer configurations instead? It's main purpose is to reflect injected values.

mviegas (commented):
Just left some initial thoughts while we mature the idea a bit more.

src/DotNetCore.CAP.AzureServiceBus/Options/AzureServiceBusOptionsBase.cs
KuchaD: Because I reuse this option every custom consumer. And AzureServiceBusOptions is extened AzureServiceBusOptionsBase for defaul options

samples/Sample.AzureServiceBus.InMemory/SampleSubscriber.cs
KuchaD: Thx I just tried it for only diffrent message. For same are exetuce both of subscriber handlers.

docs/content/user-guide/en/transport/azure-service-bus.md
```diff
@@ -109,3 +109,30 @@ c.UseAzureServiceBus(asb =>
...
+You can customize the setup for customer consumers in Azure by configuring extra service options per group. This includes integrating with extra topic or adjusting other relevant settings.
+
+```C#
+c.UseAzureServiceBus(asb =>
+{
+    asb.ConnectionString = ...
+           asb.ConfigureCustomGroupConsumer("test", cfg =>
+        {
+            cfg.UseTopic("entity-created");
+            cfg.UseConnectionString("external connection string");
```
mviegas, line 124: Do you think we should also remove this here since all consumers should be in the same namespace defined in the default config?

samples/Sample.AzureServiceBus.InMemory/Program.cs
```diff
@@ -32,9 +32,25 @@
...
         };
-
+        
         asb.ConfigureCustomProducer<EntityCreatedForIntegration>(cfg => cfg.UseTopic("entity-created").WithSubscription());
         asb.ConfigureCustomProducer<EntityDeletedForIntegration>(cfg => cfg.UseTopic("entity-deleted").WithSubscription());
+       
+        asb.ConfigureCustomTopicConsumer("test", cfg =>
+        {
+            cfg.UseTopic("entity-created");
+            cfg.UseDefaultOptions(); // Use default options from base consumer
```
mviegas, line 42: What's the effect of not calling this explicitly? Should we advise the user to do it?

src/DotNetCore.CAP.AzureServiceBus/Consumer/ServiceBusConsumerDescriptorBuilder.cs
```diff
@@ -0,0 +1,54 @@
...
+        GroupName = groupName;
+    }
+
+    public ServiceBusConsumerDescriptorBuilder UseTopic(string topicPath)
+    {
+        TopicPath = topicPath;
+        return this;
+    }
+    
+    public ServiceBusConsumerDescriptorBuilder Configuration(Action<AzureServiceBusOptionsBase> configure)
```
mviegas, line 34: What do you think of keeping the standard here and calling this method `UseConfiguration` instead?

src/DotNetCore.CAP.AzureServiceBus/Options/AzureServiceBusOptionsBase.cs
```diff
@@ -145,16 +129,5 @@ public Func<ServiceBusReceivedMessage, IServiceProvider, List<KeyValuePair<strin
     /// </summary>
     public List<KeyValuePair<string, string>>? SQLFilters { get; set; }
```
mviegas, line 130: To maintain parity, I believe we should also add this property to the custom consumer configuration, what do you think?

src/DotNetCore.CAP.AzureServiceBus/Options/AzureServiceBusOptionsBase.cs
```diff
@@ -123,11 +112,6 @@ public class AzureServiceBusOptions
     /// </remarks>
     public TimeSpan MaxAutoLockRenewalDuration { get; set; } = TimeSpan.FromMinutes(5);
```
mviegas, line 113: To maintain parity, I believe we should also add this property to the custom consumer configuration, what do you think?

src/DotNetCore.CAP.AzureServiceBus/Options/AzureServiceBusOptionsBase.cs
```diff
@@ -88,7 +77,7 @@ public class AzureServiceBusOptions
     /// Adds additional correlation properties to all correlation filters.
     /// https://learn.microsoft.com/en-us/azure/service-bus-messaging/topic-filters#correlation-filters
     /// </summary>
-    public IDictionary<string, string> DefaultCorrelationHeaders { get; } = new Dictionary<string, string>();
+    public IDictionary<string, string> DefaultCorrelationHeaders { get; set; } = new Dictionary<string, string>();
 
     /// <summary>
     /// Gets the maximum number of concurrent calls to the ProcessMessageAsync message handler the processor should initiate.
```
mviegas, line 83: To maintain parity, I believe we should also add this property to the custom consumer configuration, what do you think?

src/DotNetCore.CAP.AzureServiceBus/Consumer/ServiceBusConsumerDescriptorBuilder.cs
```diff
@@ -0,0 +1,54 @@
...
+using Azure.Core;
+
+namespace DotNetCore.CAP.AzureServiceBus.Consumer;
+
+public class ServiceBusConsumerDescriptorBuilder
+{
+    /// <summary>
+    /// Name of the group that the consumer is part of.
+    /// </summary>
+    private string GroupName { get; set; } = null!;
```
mviegas, line 15: What about renaming it to SubscriptionName to keep parity with the transport idiom?

mviegas (changes requested):
Hello! I think we're on track for a nice feature. Just did another review here with a couple of suggestions:

- Keep parity of current consumer properties also to custom consumers, with the exception of Sessions.
- Naming adjustments.

demorgi:
Have you considered just using message forwarding?
https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-auto-forwarding

You can create a single topic for your service to consume and forward messages from other topics

Issue #1583, opened 2024-09-17
Azure Service Bus Messages Marked as Completed Even When Persistence Fails


oussama-smida:
**Description**:  
In CAP 8.2, when using Azure Service Bus and SQL Server, messages are marked as completed even if the system fails to persist the received message to the database.
This issue has been observed as a rare occurrence but poses a considerable risk to the reliability of the system, as it can lead to message loss.

**Analysis**:  
The issue occurs because the default setting for `AutoCompleteMessages` in CAP is `true`. When an exception occurs during message persistence (e.g., SQL timeout), CAP attempts to call the `Reject` method. However, the current implementation of `Reject` for Azure Service Bus is a no-op, resulting in the message being automatically completed despite the failure.

**Trace Log**:
```less
An exception occurred when processing the received message. Message:'{0}'.
Source context: DotNetCore.CAP.Internal.ConsumerRegister
```
**Exception**:
```less
Microsoft.Data.SqlClient.SqlException (0x80131904): Execution Timeout Expired. The timeout period elapsed prior to completion of the operation or the server is not responding.
 ---> System.ComponentModel.Win32Exception (258): Unknown error 258
   at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
...
```
**Sql Command**:
```sql
INSERT INTO cap.Received([Id],[Version],[Name],[Group],[Content],[Retries],[Added],[ExpiresAt],[StatusName]) VALUES(@Id,'v1',@Name,@Group,@Content,@Retries,@Added,@ExpiresAt,@StatusName);
```

**Relevant source code**
DotNetCore.CAP.Internal.ConsumerRegister
```csharp
try
{
...
}
catch (Exception e)
{
    _logger.LogError(e, "An exception occurred when process received message. Message:'{0}'.",
        transportMessage);

    client.Reject(sender); // NO-OP in the Azure Service Bus IConsumerClient implementation

    TracingError(tracingTimestamp, transportMessage, client.BrokerAddress, e);
}
```
**Proposed Fix**:
DotNetCore.CAP.AzureServiceBus.AzureServiceBusConsumerClient
```csharp
public void Reject(object? sender)
{
    var commitInput = (AzureServiceBusConsumerCommitInput)sender!;
    commitInput.AbandonMessageAsync().GetAwaiter().GetResult();
}
```

yang-xiaodong:
[...]

PR #1584, opened 2024-09-17
fix(#1583): Call AbandonMessageAsync on Azure Service Bus message when persistence fails to avoid message loss

oussama-smida:
### Description:
When using Azure Service Bus, messages are marked as completed even if the system fails to persist the received message to the database.

#### Issue(s) addressed:
- [Azure Service Bus Messages Marked as Completed Even When Persistence Fails](https://github.com/dotnetcore/CAP/issues/1583)

#### Changes:
- Call AbandonMessageAsync on Azure Service Bus message when persistence fails to avoid message loss.

#### Affected components:
- DotNetCore.CAP.AzureServiceBus

### Checklist:

### DIFF
--- a/src/DotNetCore.CAP.AzureServiceBus/AzureServiceBusConsumerClient.cs
+++ b/src/DotNetCore.CAP.AzureServiceBus/AzureServiceBusConsumerClient.cs
@@ -134,7 +134,8 @@ internal sealed class AzureServiceBusConsumerClient : IConsumerClient
 
     public void Reject(object? sender)
     {
-        // ignore
+        var commitInput = (AzureServiceBusConsumerCommitInput)sender!;
+        commitInput.AbandonMessageAsync().GetAwaiter().GetResult();
     }
 
     public void Dispose()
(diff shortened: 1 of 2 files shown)


mviegas:
[MS Docs](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusprocessoroptions.autocompletemessages?view=azure-dotnet) have the following remark about the `AutoCompleteMessages` property of the SB processor:

> If the message handler triggers an exception and did not settle the message, then the message will be automatically abandoned, irrespective of [AutoCompleteMessages](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusprocessoroptions.autocompletemessages?view=azure-dotnet#azure-messaging-servicebus-servicebusprocessoroptions-autocompletemessages).

Still, what you proposed here makes sense as the `OnMessageCallback` catches exceptions, which causes the processor to think that messages were handled successfully.

LGTM.

mviegas (approved):

PR #1585, opened 2024-09-17
added control for queue options in RabbitMQ transport

apatozi:
### Description:
I have a use case where transient services that are non competing consumers (consumer group GUID) have to use the queue system to get some push events from other services. These queues are created but never deleted when the service stops to come back in a later time. When the service gets back up it creates a new GUID consumer group, so a new queue. For the moment I am manually deleting these using the RabbitMQ management API so each service when starts, checks for queues with 0 consumers. 
My actual solution is hacky and doesn't work every time. 
_With my ugly hacky implementation there is a problem because sometimes when 2 services start up at the same time, each create its queue but the consumers threads haven't connected yet so the queue gets deleted by the other service. This could have a beautiful solution by using the "Auto Delete" feature of RabbitMQ queues so each queue deletes itself if looses all the consumers._ 
I needed to control the queues auto-delete options so I made this PR 

#### Issue(s) addressed:
- did not open issue 

#### Changes:
- added QueueOptions to RabbitMQOptions so you can control AutoDelete of a queue.
- allowed to change also durable and exclusive options  for any future use case 
- left default options as they were 

#### Affected components:
- DotNetCore.CAP.RabbitMQ

#### How to test:
create a rabbitMQ with new options and check those in the management 

### Checklist:
- [x] I have updated the relevant tests (if applicable) [Not applicable]

### Reviewers:

### DIFF
--- a/docs/content/user-guide/en/transport/rabbitmq.md
+++ b/docs/content/user-guide/en/transport/rabbitmq.md
@@ -50,6 +50,7 @@ VirtualHost | Broker virtual host | string | /
 Port | Port | int | -1
 ExchangeName | Default exchange name | string | cap.default.topic
 QueueArguments  | Extra queue `x-arguments` | QueueArgumentsOptions  |  N/A
+QueueOptions  | Change Options for created queue | QueueRabbitOptions  |  { Durable=true, Exclusive=false, AutoDelete=false }
 ConnectionFactoryOptions  |  RabbitMQClient native connection options | ConnectionFactory | N/A
 CustomHeadersBuilder  | Custom subscribe headers |  See the blow |  N/A
 PublishConfirms | Enable [publish confirms](https://www.rabbitmq.com/confirms.html#publisher-confirms) | bool | false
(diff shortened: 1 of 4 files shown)


yang-xiaodong:
Hi,
I have a question. Our consumer group name is declared using an Attribute, which means it cannot be specified dynamically at runtime (dynamic consumer group GUID). How did you do to achieve that?

apatozi:
> Hi, I have a question. Our consumer group name is declared using an Attribute, which means it cannot be specified dynamically at runtime (dynamic consumer group GUID). How did you do to achieve that?

I achieve that by using the DefaultGroupName in the AddCap() method. 

```
services.AddCap(x =>{
    x.DefaultGroupName = CapInstance.InstanceId.ToString();
    x.GroupNamePrefix = $"{capSettings.GroupNamePrefix}.{capSettings.DefaultGroupName}";
});
```

This is used in a WebSocket BFF and events are used to push messages from other services to the clients passing via the BFF. As there are multiple instances of this BFF service, each message should be delivered to every instance because  (for ex) the Notification service doesn't know which service has a specific WS client connection.

yang-xiaodong:
Thank you for your explanation. The current options seem to fit your use case very well. I'm willing to merge it in, thank you for your PR!

yang-xiaodong:
Fixed in version `8.3.0-preview-243613753`

Issue #1626, opened 2024-12-20
Incorrect order of produced messages in transaction


PoteRii:
Hello,
Consider a scenario (slightly modified code from sample `Sample.Kafka.PostgreSql` project):
![image](https://github.com/user-attachments/assets/468aabea-50eb-4373-bfd2-07f9cfc6d517)

Sometimes we are seeing that produced messages are in incorrect order, meaning that message with less `DateTime.Now` value has higher offset

![image](https://github.com/user-attachments/assets/f6ad25a2-ed81-4c9f-9581-f774d9dca680)

this happens not always, but quite frequently.
In our scenario even adding `await Task.Delay(500);` between two messages does not help.

Could it be because there is no await in `Flush` method? In that case possible solution might be new `FlushAsync` function

yang-xiaodong:
Does using `PublishAsync` have the same issue?

PoteRii:
Yes. In our real scenario we use `PublishAsync`.

PoteRii:
I`ve tested by adding `FlushAsync` function and doing `await FlushAsync` in `CommitAsync` and it seems to work

PoteRii:
should I make a PR with new `FlushAsync`? At least it won`t break anything

yang-xiaodong:
> should I make a PR with new `FlushAsync`? At least it won`t break anything

Sure, Thanks

PoteRii:
https://github.com/dotnetcore/CAP/pull/1629

yang-xiaodong:
Fixed in version 8.3.3-preview-255432523

PR #1629, opened 2024-12-26
Added `FlushAsync` function which is called from `CommitAsync` with await keyword

PoteRii:
### Description:
Added `FlushAsync` function which is called from `CommitAsync` with await keyword to ensure that multiple messages are produced in correct order from a single transaction

#### Issue(s) addressed:
- https://github.com/dotnetcore/CAP/issues/1626

#### How to test:
Produce multiple messages inside a single transaction and check the order

### Checklist:
- [ ✓] I have tested my changes locally
- [ ✓] My changes follow the project's code style guidelines

### DIFF
--- a/src/DotNetCore.CAP/ICapTransaction.Base.cs
+++ b/src/DotNetCore.CAP/ICapTransaction.Base.cs
@@ -68,6 +68,29 @@ public abstract class CapTransactionBase : ICapTransaction
         }
     }
 
+    protected virtual async Task FlushAsync()
+    {
+        while (!_bufferList.IsEmpty)
+        {
+#pragma warning disable CA2012 // Use ValueTasks correctly
+            if (_bufferList.TryDequeue(out var message))
+            {
+                var isDelayMessage = message.Origin.Headers.ContainsKey(Headers.DelayTime);
+                if (isDelayMessage)
+                {
+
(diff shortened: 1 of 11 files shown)


src/DotNetCore.CAP/ICapTransaction.Base.cs
```diff
@@ -68,6 +68,29 @@ protected virtual void Flush()
         }
     }
 
+    protected virtual async Task FlushAsync()
```
yang-xiaodong, line 71: Why not have Flush() call FlushAsync()?

src/DotNetCore.CAP/ICapTransaction.Base.cs
```diff
@@ -68,6 +68,29 @@ protected virtual void Flush()
...
+    {
+        while (!_bufferList.IsEmpty)
+        {
+#pragma warning disable CA2012 // Use ValueTasks correctly
```
yang-xiaodong, line 75: Remove the `#pragma warning`

yang-xiaodong (changes requested):

src/DotNetCore.CAP/ICapTransaction.Base.cs
PoteRii: removed

src/DotNetCore.CAP/ICapTransaction.Base.cs
PoteRii: changed

yang-xiaodong (approved):
LGTM, Thanks

PR #1638, opened 2025-01-14
Add thread-safe message scheduling and related tests

### LINKED ISSUE #1637: Thread-safety and cancellation handling issues in PublishDelayAsync with delay time < 1 minute
Version: 8.3.2

### **Problem Summary**
The `ICapPublisher.PublishDelayAsync` method in the .NET Core CAP library exhibits thread-safety and cancellation handling issues when the delay time is less than 1 minute. These issues can lead to silent failures and message accumulation in the database.

---

### **Detailed Problems**

- **Thread-safety Issue:**
  - The `Dispatcher.EnqueueToScheduler` method uses a non-thread-safe `PriorityQueue` to manage delayed messages.
  - Under race conditions, the `PriorityQueue` can return `null` messages. 
  - This causes the fire-and-forget task (Dispatcher: line 80) to fail silently, with no error logging.
  - As a result, delayed message processing halts, and future scheduled messages are not sent until the application restarts.

- **Cancellation Handling Issue:**
  - The `_delayCts` cancellation token is initialized only once.
  - Once `_delayCts` is cancelled, subsequent calls to `await Task.Delay(new TimeSpan(delayTime), _delayCts.Token)` throw an `OperationCanceledException`.
  - The exception is ignored in the `catch` block, leading to a `while (TryPeek)` loop that spins continuously until the delay time expires.

- **Consequences of Silent Failures:**
  - Delayed messages continue to accumulate in the database without being processed.
  - Upon application restart, querying delayed messages without a limit causes the database to timeout.
  - Manual intervention is required to delete unprocessed messages and restore application functionality.

---

### **Steps to Reproduce**

1. Call `ICapPublisher.PublishDelayAsync` concurrently with delay times less than 1 minute.
2. Observe that:
   - Some messages are not processed.
   - Future scheduled messages are not sent.
---

I have fixed the above problems while maintaining existing functionality as closely as possible. I will open a pull request with tests and code changes to address the issue.

amimelia:
Introduce `ScheduledMediumMessageQueue` for thread-safe scheduling of messages. Updated `Dispatcher` to use the new queue and modified the scheduling logic for improved reliability. Added unit tests.

Fixes #1637

### DIFF
--- a/src/DotNetCore.CAP/Internal/ScheduledMediumMessageQueue.cs
+++ b/src/DotNetCore.CAP/Internal/ScheduledMediumMessageQueue.cs
@@ -1,0 +1,88 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using DotNetCore.CAP.Persistence;
+
+namespace DotNetCore.CAP.Internal;
+
+public class ScheduledMediumMessageQueue
+{
+    private readonly SortedSet<(long, MediumMessage)> _queue = new(Comparer<(long, MediumMessage)>.Create((a, b) =>
+    {
+        int result = a.Item1.CompareTo(b.Item1);
+        return result == 0 ? String.Compare(a.Item2.DbId, b.Item2.DbId, StringComparison.Ordinal) : result;
+    }));
+
+    private readonly SemaphoreSlim _semaphore = new(0);
+    private readonly object _lock = new();
+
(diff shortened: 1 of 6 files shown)


yang-xiaodong (approved):
LGMT, Thanks

PoteRii:
kudos for using `SortedSet`, because a message can end up in `PriorityQueue` twice if it`s not processed fast enough. Because one can have many delayed messages to be sent at the same time and job runs every 1 min so next job might pick up Queued messages that are still in `PriorityQueue`.

PR #1641, opened 2025-01-17
refresh data after re-executed items

maxwell60701:
### Description:
refresh data after re execute


yang-xiaodong (approved):
LGTM, Thanks

PR #1664, opened 2025-03-29
Modify document general. md

findersky:
### Description:
Add the implementation project address for the Dameng database in the document to save more time for projects that require Dameng database support.

#### Changes:
-  docs/content/user-guide/zh/storage/general.md


yang-xiaodong (approved):
Thanks

PR #1680, opened 2025-04-28
feat: cap subscribe existing ef transaction

Badea741:
### Description:
This PR will add a new function in SQL server transaction that makes the user be able to make cap publisher subscribe to an existing entity framework transaction which has a good motivation.
The motivation is to let the user begin a scoped transaction anywhere in the request and let cap subscribe to this transaction later on this is beneficial when having an interceptor on entity framework save changes which may have an enclosing transaction, as a result of entity framework and SQL doesn't support nested transaction so you have to reuse the same transaction that is opened in the scoped context

#### Issue(s) addressed:
- no issue addressed

#### Changes:
- Added an extension method in ICapTransaction.SqlServer named SubscribeCap on the 'IDbContextTransaction'
- Added an endpoint in the ValuesController which is included in SQL Server, RabbitMQ sample named 'EntityFrameworkWithExistingTransaction'

#### Affected components:
- Sample.RabbitMQ.SqlServer.Controllers.ValuesController
- ICapTransaction.SqlServer

#### How to test:
I've included a sample in the SQL server, RabbitMQ sample which can be helpful in testing it, it's routed at '/ef/transaction/existing'

### Checklist:

### DIFF
--- a/samples/Sample.RabbitMQ.SqlServer/Controllers/ValuesController.cs
+++ b/samples/Sample.RabbitMQ.SqlServer/Controllers/ValuesController.cs
@@ -1,12 +1,12 @@
-﻿using System;
-using System.Data.Common;
-using System.Threading.Tasks;
-using Dapper;
+﻿using Dapper;
 using DotNetCore.CAP;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Data.SqlClient;
 using NameGenerator.Generators;
 using Sample.RabbitMQ.SqlServer.Messages;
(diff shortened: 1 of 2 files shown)


yang-xiaodong:
Thank you for your PR. It doesn't need to be added to the library code. This is user code. We need to be consistent among different providers.

xiangxiren:
> Thank you for your PR. It doesn't need to be added to the library code. This is user code. We need to be consistent among different providers.

Agree!

PR #1727, opened 2025-08-21
Fix RabbitMQ consuming body reusing

otemnov:
### Description:
In case of high concurrency, the consumer now uses `ArrayPool<byte>.Shared` to copy the delivery body into a safe buffer before dispatching work to background tasks. It is required according to RabbitMQ documentation.
This avoids accessing memory that may be released by RabbitMQ after `HandleBasicDeliverAsync` returns and reduces allocations under heavy load.  

#### Issue(s) addressed:
- https://github.com/dotnetcore/CAP/issues/1728 

#### Changes:
- Copying message body immediately at the beginning of `HandleBasicDeliverAsync`.  
- Using `ArrayPool<byte>.Shared` instead of allocating a new array with `ToArray()`.  
- Ensuring buffer is safely returned to the pool in a `finally` block.  
- Updated `Consume` to work with `ReadOnlyMemory<byte>` built from the rented buffer.  

#### Affected components:
- Message consumer implementation (`HandleBasicDeliverAsync` logic).  
- Downstream message handling that consumes `TransportMessage`.  

#### How to test:
1. Run a RabbitMQ producer that sends a high volume of messages (e.g., 10k+ msgs).  
2. Start the consumer with `_usingTaskRun = true` to enable concurrent background processing.  
3. Verify:  
   - No exceptions occur related to accessing released memory.  
   - Buffers are returned to the pool (memory usage should stay stable under load).  
   - Message headers and body are delivered correctly to `_msgCallback`.  
4. Optionally, run performance profiling:
   - Compare GC pressure and allocations with/without `ArrayPool`.  

### Checklist:

### DIFF
--- a/src/DotNetCore.CAP.RabbitMQ/RabbitMQBasicConsumer.cs
+++ b/src/DotNetCore.CAP.RabbitMQ/RabbitMQBasicConsumer.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Buffers;
 using System.Collections.Generic;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


yang-xiaodong:
Hi

Thanks for the PR, Using `ArrayPool<byte>` here is a reasonable optimization, but not strictly necessary. The main benefit is reducing GC pressure in very high-throughput or large-message scenarios. However, it adds complexity: the buffer lifetime must be carefully managed and always returned, otherwise it risks corruption or leaks.


```
var safeBody = _usingTaskRun ? body.ToArray() : body;
```

For most scenarios, `ToArray()` is simpler and safer. The `ArrayPool` approach only makes sense if memory allocation and GC overhead have been proven to be a real performance bottleneck.

otemnov:
@yang-xiaodong Hi, sure I've switched back to simple array copy

Issue #1738, opened 2025-09-11
Disable Kafka Automatic Topic Creation


kl-tie:
## Background
Currently we want to make a comprehensive control for our Kafka cluster, like topic creation policy by using SASL authentication to authenticate with our Kafka cluster. CAP does create the topic automatically by default. AFAIK, you should be able to disable this behavior by setting the librdkafka `allow.auto.create.topics` to `false`.

## Issue
So, we have changed the `allow.auto.create.topics` like this in our app:
```csharp
builder.Services.AddCap(x =>
{
    x.UseInMemoryController();
    x.UseKafka(kafkaOptions =>
    {
        kafkaOptions.MainConfig.Add("allow.auto.create.topics", "false");
        kafkaOptions.MainConfig.Add("security.protocol", "sasl_plaintext");
        kafkaOptions.MainConfig.Add("sasl.mechanism", "SCRAM-SHA-256");
        kafkaOptions.MainConfig.Add("sasl.username", "dotnet-app-user");
        kafkaOptions.MainConfig.Add("sasl.password", "helloworld");
        kafkaOptions.Servers = "localhost:9092";
    });   
});
```

And with this consumer:
```csharp
[ApiController]
[Route("[controller]")]
public class ConsumerController : ControllerBase
{
    private readonly ILogger<ConsumerController> _logger;

    public ConsumerController(ILogger<ConsumerController> logger)
    {
        _logger = logger;
    }

    [NonAction]
    [CapSubscribe("dotnet-test")]
    public void ReceiveMessage(string message)
    {
        _logger.LogInformation("Received message: {Message}", message);
    }
}
```

But it still trying to create the topics automatically with these exceptions:
```
fail: DotNetCore.CAP.Internal.ConsumerRegister[0]
      Kafka client consume error. --> An error was encountered when automatically creating topic! -->An error occurred creating topics: [dotnet-test]: [Unauthorized].
info: DotNetCore.CAP.Processor.CapProcessingServer[0]
      Starting the processing server.
fail: DotNetCore.CAP.Internal.ConsumerRegister[0]
      FindCoordinator response error: Broker: Group authorization failed
      Confluent.Kafka.ConsumeException: FindCoordinator response error: Broker: Group authorization failed
         at Confluent.Kafka.Consumer`2.Consume(Int32 millisecondsTimeout)
         at Confluent.Kafka.Consumer`2.Consume(TimeSpan timeout)
[...]

PR #1741, opened 2025-09-15
Honor allow.auto.create.topics for Kafka

yang-xiaodong:
## Summary
- respect allow.auto.create.topics and skip topic creation when disabled
- document how to turn off Kafka topic auto-creation

## Testing
- `dotnet test` *(fails: DotNetCore.CAP.MySql.Test.*)*

------

Issue: #1738

### DIFF
--- a/docs/content/user-guide/en/transport/kafka.md
+++ b/docs/content/user-guide/en/transport/kafka.md
@@ -64,6 +64,18 @@ services.AddCap(capOptions =>
 
 [https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md)
 
+To prevent CAP from creating topics automatically, disable topic auto creation:
+
+```csharp
+services.AddCap(capOptions =>
+{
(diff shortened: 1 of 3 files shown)


kl-tie:
@yang-xiaodong currently my team is still using CAP v7, will you accept a PR from me that use the latest v7 with the similar changes from this PR?

yang-xiaodong:
> @yang-xiaodong currently my team is still using CAP v7, will you accept a PR from me that use the latest v7 with the similar changes from this PR?

No, we won't release any old version, even if we have bugs. This is an open source project and does not have complete commercial support.

kl-tie:
> This is an open source project and does not have complete commercial support.

That's why I'm offering the support for older versions by making a PR for the previous version, especially since the changes are non-breaking.

yang-xiaodong:
This has nothing to do with who makes PR. This is our release policy.  By the way, version 8.x is compatible with the feature of 7.x.

kl-tie:
Ok then, we will try to upgrade the CAP version to 8.x first.

> This is an open source project and does not have complete commercial support.

BTW, you did mention about commercial support on this statement, does CAP actually have some kind of commercial support?

PR #1761, opened 2025-10-29
eat(cap): support configurable message expiration time

guoxdxdxd:
• Added SucceedMessageExpiredAfter and FailedMessageExpiredAfter properties to TopicAttribute for configuring expiration times of successful and failed messages.
• Updated ISubscribeExecutor implementation to apply expiration times based on topic attribute settings.
• Modified method signatures of SetSuccessfulState and SetFailedState to include expiration time parameters.
• Optimized message expiration logic to prioritize configuration values from topic attributes; fallback to global defaults if not specified.

Description:

In some scenarios, it may be necessary to customize the expiration time for a specific topic. For example, less important messages in large quantities may have shorter expiration times, while very important messages may require longer expiration times. Configuring expiration times per individual topic allows for more flexibility to meet user requirements.

Changes:
	•	Added expiration time settings in TopicAttribute.
	•	Updated SubscribeExecutor to retrieve custom configuration.

Checklist:

### DIFF
--- a/src/DotNetCore.CAP/Internal/ISubscribeExector.Default.cs
+++ b/src/DotNetCore.CAP/Internal/ISubscribeExector.Default.cs
@@ -60,7 +60,7 @@ internal class SubscribeExecutor : ISubscribeExecutor
                 TracingError(DateTimeOffset.UtcNow.ToUnixTimeMilliseconds(), message.Origin, null, new Exception(error));
 
                 var ex = new SubscriberNotFoundException(error);
-                await SetFailedState(message, ex);
+                await SetFailedState(message, ex, descriptor.Attribute.FailedMessageExpiredAfter);
                 return OperateResult.Failed(ex);
             }
         }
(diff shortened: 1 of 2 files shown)


yang-xiaodong:
Sorry, this feature was not approved because I don't consider it a valid or common requirement. For your needs, could you set a maximum value and then start a background thread to delete some messages?

guoxdxdxd:
> Sorry, this feature was not approved because I don't consider it a valid or common requirement. For your needs, could you set a maximum value and then start a background thread to delete some messages?

I believe that data created by CAP should be maintained as much as possible within CAP itself.
In my current scenario, which is relatively complex, maintaining it via attributes seems to be the least costly approach.
Alternatively, we might want to reconsider whether this requirement is actually reasonable.

demorgi:
You can use the consumer filter to decide not to process the message based on your logic.
https://cap.dotnetcore.xyz/user-guide/en/cap/filter/

Issue #1766, opened 2025-11-11
Regression: Fix RabbitMQ consuming body reusing


otemnov:
### Summary

Regression: Fix RabbitMQ consuming body reusing 

### Steps to Reproduce

The following fix https://github.com/dotnetcore/CAP/pull/1727 is broken by the change 
https://github.com/dotnetcore/CAP/commit/273fe35ff4486dfe4c7f1cca170fd72940ec6a27

### Expected Behavior

safeBody should be used, not unsafe body 

### Log Output

```shell

```

### CAP Configuration

```yaml

```

### Transport Used

RabbitMQ

### Storage Provider

None

yang-xiaodong:
Fixed in version 10

PR #1767, opened 2025-11-11
Fix RabbitMQ consumer concurrency

otemnov:
- Fixes [Regression: Fix RabbitMQ consuming body reusing](https://github.com/dotnetcore/CAP/issues/1766)
- Moved the "Consume" local function to a private method to make it clearer when and what arguments we pass.


otemnov:
@yang-xiaodong could your assign reviewer?

PR #1790, opened 2026-02-12
Optimize transaction cleanup to prevent memory leaks

wificlub-coder:
Enhanced the setter of the Transaction property in CapPublisher: when the property is assigned a null value, _asyncLocal.Value is cleared to avoid memory leaks caused by ExecutionContext holding a reference to CapTransactionHolder. Additionally, Transaction = null was added to the finally block of the method to ensure timely release of transactions.

### Description:
_Provide a more detailed description of the changes made in this PR. Explain the reason behind the changes and their expected impact on the project._

#### Issue(s) addressed:
- [Link to the related issue(s), if any]

#### Changes:
- _List the major changes made in this PR, preferably in bullet points._

#### Affected components:
- _List the components of the project that are affected by this PR._

#### How to test:
_Provide a step-by-step guide on how to test your changes. Include any relevant information like environment setup, dependencies, etc._

### Checklist:
- [ ] I have tested my changes locally

### Reviewers:

### DIFF
--- a/src/DotNetCore.CAP/Internal/ICapPublisher.Default.cs
+++ b/src/DotNetCore.CAP/Internal/ICapPublisher.Default.cs
@@ -46,6 +46,12 @@ internal class CapPublisher : ICapPublisher
         get => _asyncLocal.Value?.Transaction;
         set
         {
+            if (value == null)
+            {
+                // Clear the CapTransactionHolder held in the ExecutionContext to avoid long-term reference retention.
+                _asyncLocal.Value = null;
+                return;
+            }
             _asyncLocal.Value ??= new CapTransactionHolder();
             _asyncLocal.Value.Transaction = value;
         }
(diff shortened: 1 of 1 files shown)


yang-xiaodong:
Thanks for the PR. I don’t think this is a memory leak in the current implementation. AsyncLocal keeps the CAP transaction in the current execution context and does not grow as a collection. The transaction resource is expected to be released by the transaction lifecycle (Commit/Rollback/Dispose), and SQL Server transaction buffers are also removed on commit/rollback/connection-close diagnostic events.
Clearing Transaction in PublishInternalAsync finally may break scenarios where multiple messages are published within the same transaction. After the first publish, the ambient CAP transaction would be cleared, and subsequent publishes could be stored outside the intended database transaction.
If there is a reproducible leak, please provide a minimal repro and memory profile. Otherwise, I don’t think this change should be accepted.

